Issue #5, opened 2022-04-04
Just pass through value when no format is specified 


statasaurus:
At the moment if there is a row that isn't covered by a format it will just fall over

thebioengineer:
Should we by default provide a fmt_str(.default, .default, .default, fmt()) element style?

PR #10, opened 2022-04-06
Param update

statasaurus:


statasaurus:
Also fixes #5

R/theme_fmt.R
```diff
@@ -25,7 +25,11 @@ fmt_combine <- function(expression, ..., missing = NULL){
 
 
 
-fmt_str <- function(group = NULL, label = NULL, fmt = NULL){
+fmt_str <- function(group = NULL, label = NULL, ...){
+  fmt <- list(...)
+  if(length(fmt) > 1){
```
bzkrouse, line 30: Do we have plans to accommodate multiple params in 1 fmt_str()? for example 
```
fmt_str(group = ".default", label = ".default", n = fmt("XXX"), pct = fmt("xx.x%"))
```

statasaurus: At the moment no, because I think it will be confusing on the user end. But, we could, there isn't a technical reason not to. I am probably going to leave it as 1 now, just cause it is clear that fmt_str only ever applies one format either of format combine or just a normal format. If people don't find this confusing and find it more annoying to re-specify the group and label then we can change it

R/utils.R
```diff
@@ -144,46 +152,90 @@ apply_combo_fmt <- function(.data, fmt_combine, param, values){
 #'
 #' @noRd
 #' @importFrom rlang as_label
-param_test_expr <- function(param, val){
+expr_to_filter <- function(param, val){
```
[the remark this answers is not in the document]
statasaurus: Fair I named it param cause it is the parameter group or label. If you have a better name I am happy to go with it (I am not that creative). It is just an internal function so I am not super worried about it. But your point is fair

PR #27, opened 2022-04-13
rename to frmt

thebioengineer:
Hey, i was taking a look at some of the logic in the apply_tfmt function, and ended up doing a bit of exploration. I've ended up renaming as we talked about to use frmt instead of fmt in most location and element_style to table_body_plan. I also added print/format functions to frmt, frmt_structure, frmt_combine, table_body_plan. 

Finally, I tried to break down some of the logic into smaller parts where the formatting was being applied, but woof, there is so much there. anyway, I wanted to get this in to chat about.


statasaurus:
So cause this is all one commit I can't really cherry-pick through it. I am just going to go through and try to copy over some stuff. But you made a lot of changes

PR #30, opened 2022-04-13
Add examples

### LINKED ISSUE #19: apply_fmt throws error in the case of bounds formatting and NA values
error: Error in apply_fmt(vals, fmt) : 
  Check format, there largest value is larger than expected

to reproduce: 

```
fmt <- list(rounding = "x.xxx", bounds = element_bounds(upper_exp = ">0.99", upper_lab = ">0.99", lower_exp = "<0.001", lower_lab = "<0.001"))
vals <- c(NA, 0.0547, 0.9999)
apply_fmt(vals, fmt)
```
In this case, the calculation of space_to_add in fmt_options becomes negative

bzkrouse:
- Add demog & AE examples
  - @statasaurus as we discussed there is an issue with the demog tbl fmt_combine. I tried to dig into this and noticed that indices were being represented in multiple formats (specifically the fmt_combine and the fmt for the p-value)
 
- also resolves #19 
- fix small issue in the building of grp expression in fmt_test_data in case of multiple groups

### DIFF
--- a/R/theme_element.R
+++ b/R/theme_element.R
@@ -56,7 +56,7 @@ element_bounds <- function(upper_exp = NULL, lower_exp = NULL, upper_lab = NULL,
     upper_lab = upper_exp
   }
 
-  if(!is_null(lower_exp) & is_null(lower_lab)){
+  if(!is.null(lower_exp) & is.null(lower_lab)){
     lower_lab = lower_exp
   }
   structure(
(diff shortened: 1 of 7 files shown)


statasaurus:
Move to go into plan_apply_cleaned rather than into main so it can get the updated code. plan_apply_cleaned will go into main once frmt_when and the method functions for apply format are done

PR #40, opened 2022-04-19
Improved layering - quos and vars

thebioengineer:
Update layering to prevent a tfrmt from selecting an empty quo/vars over a populated one.


statasaurus (approved):
Looks great thanks for adding this

statasaurus:
@thebioengineer So I have changed the names of the functions to match with the frmt. Can you just check this all still works right?

thebioengineer:
It all looks to work. Had to update some of the tests that were referencing tfmt not tfrmt. That is now updated and should be all good now.

PR #42, opened 2022-04-19
add element_align functionality

### LINKED ISSUE #33: Align column options

bzkrouse:
resolves #33 

@statasaurus @thebioengineer For the decimal alignment, i am aligning on decimal, space & comma under the hood. Curious your thoughts on generalizing this to be a character alignment w/ user-specified regex


statasaurus (changes requested):
I don't think we want to remove white space when it is left aligned

bzkrouse:
> I don't think we want to remove white space when it is left aligned

@statasaurus this has been updated!

bzkrouse:
@statasaurus Updated to preserve leading white spaces. Also generalized the dec alignment to be character alignment. We can keep thinking on naming convention of the args (left & right vs. char & char_val)

R/col_align.R
```diff
@@ -0,0 +1,74 @@
...
+
+#' Align value columns based on element_align specs
+#'
+#' @param .data data
+#' @param col_align element_align object
+#'
+col_align_all <- function(.data, col_align){
+
+  .data %>%
+    mutate(across(col_align$char %>% map(quo_name) %>% unlist, ~col_align_char(.x, char_val = col_align$char_val)),
```
statasaurus, line 70: This isn't super important, but I usually just use map_chr rather than map and unlist. I find it a bit easier to read, but feel free to ignore 

```suggestion
    mutate(across(col_align$char %>% map_chr(quo_name), ~col_align_char(.x, char_val = col_align$char_val)),
```

statasaurus (approved):
Looks great! Can you just resolve the merge conflicts and then we can get it pulled into main 
🥳

PR #49, opened 2022-04-22
Frmt when

statasaurus:


R/apply_tfrmt_methods.R
```diff
@@ -0,0 +1,240 @@
...
+#' @importFrom rlang :=
+apply_frmt.frmt_when <- function(frmt_def, .data, values, ...){
+  values_str <- as_label(values)
+  n <- length(frmt_def)
+
+  val_len <- length(pull(.data, !!values))
+  right <- frmt_def %>%
+    map(f_rhs) %>%
+    map(function(x) {
+      if("frmt" %in% class(x)){
```
thebioengineer, line 211:
```suggestion
      if(is_frmt(x)){
```
Maybe simpler

R/apply_tfrmt_methods.R
```diff
@@ -0,0 +1,240 @@
...
+    map(eval_tidy, .data)
+
+
+  out <- rep(NA_character_, val_len)
+  replaced <- rep(FALSE, val_len)
+
+  for(i in seq_len(n)){
+    out <- replace_val(out, left[[i]] & !replaced, right[[i]])
+    replaced <- replaced | (left[[i]] & !is.na(left[[i]]))
+  }
```
thebioengineer, line 232: Would it be less computationally expensive to sort this out first? then only apply the frmt's as necessary? Just a discussion point, not a need to change IMO. This seems to be how ifelse and case_when handles it usually - calculate the values and then drop them into a vector of expected outputs.

R/utils_tfrmt.R
```diff
@@ -44,3 +43,40 @@ tentative_process <- function(.data, fx, param){
...
+#'
+#' @param .data Data to apply the tfrmt to
+#' @param tfrmt tfrmt object to apply to the data
+#'
+#' @return formatted tibble
+#' @noRd
+#' @importFrom purrr map safely
+#' @importFrom rlang !! !!!
+#' @importFrom dplyr select
+validat_cols_match <- function(.data, tfrmt){
```
thebioengineer, line 58:
```suggestion
validate_cols_match <- function(.data, tfrmt){
```

R/utils_tfrmt.R
```diff
@@ -1,27 +1,26 @@
...
 #'
-#' @return
-#' @export
-#'
-#' @examples
-apply_tfrmt <- function(.data, tfrmt_spec){
+#' @return formatted tibble
+#' @noRd
+apply_tfrmt <- function(.data, tfrmt){
+  validat_cols_match(.data, tfrmt)
```
thebioengineer, line 9:
```suggestion
  validate_cols_match(.data, tfrmt)
```

dev/mock_tbl.R
```diff
@@ -29,7 +29,10 @@ tfrmt_spec  <- tfrmt(
                    frmt_combine(
                      "{count} {percent}",
                      count = frmt("XXX"),
-                     percent = frmt("(XXX.X%)")
+                     percent = frmt_when("==100"~ frmt(""),
```
thebioengineer, line 32: how is frmt("") different from ""?

tests/testthat/test-utils_tfrmt.R
```diff
@@ -0,0 +1,110 @@
...
+    apply_tfrmt(raw_dat, tfrmt(
+      label = "label2")),
+      "Variable Specified in 'label' doesn't exist in the supplied dataset. Please check the tfrmt and try again."
+  )
+
+  expect_error(
+    apply_tfrmt(raw_dat, tfrmt(
+      group = "label2")),
+      "Variable Specified in 'group' doesn't exist in the supplied dataset. Please check the tfrmt and try again."
+    )
```
thebioengineer, line 107: IMO these should be in a separate test. Not a hill I'll die on though.

thebioengineer (approved):
No show-stopping issues here I think. Left comments for discussion/suggestion. 

We probably want to add a format/print method for frmt_when. But that doesn't have to stop this PR IMO

bzkrouse (approved):
Amazing progress!! ⭐  🤩 
A few suggestions:
- frmt_structure() could use some improved documentation since it's such an important piece
- Create a format/print method for frmt_when objects
- rename apply_tfrmt_methods.R to apply_frmt_methods.R
- Thoughts on using inheritParams for the repeated params like group, label, etc?

R/apply_tfrmt_methods.R
statasaurus: I don't think there is much of a cost difference, given calculating frmt's is relatively cheep. But, yes we can keep this tag for refactor if speed becomes an issue

R/utils_tfrmt.R
statasaurus: Spelling is overrated 🙄

dev/mock_tbl.R
statasaurus: It isn't really. But I think it is reasonable that people will use both options so I want to make sure both work

tests/testthat/test-utils_tfrmt.R
statasaurus: 🤷 Up to you. I probably won't change it but also I won't be mad if you change it

statasaurus:
> Amazing progress!! ⭐ 🤩 A few suggestions:
> 
> * frmt_structure() could use some improved documentation since it's such an important piece
> * Create a format/print method for frmt_when objects
> * rename apply_tfrmt_methods.R to apply_frmt_methods.R
> * Thoughts on using inheritParams for the repeated params like group, label, etc?

I totally agree with the top three comments and that would be the plan. For the inheritParams I haven't really used them much in the past. I have nothing against them though. For the methods functions I think we probably just want to make a single user doc. Or, maybe not if they are all part of the same family then you could still easily connect all of them.... Don't know need to discuss more

R/apply_tfrmt_methods.R
```diff
@@ -0,0 +1,240 @@
...
+#'
+#'
+NULL
+
+
+#' Apply formatting
+#'
+#' @param .data data, but only what is getting changed
+#' @param frmt_def formatting to be applied
+#' @param values values symbol should only be one
```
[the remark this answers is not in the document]
statasaurus: Yeah we should probably move the order around I was just lazy 💤

R/utils_tfrmt.R
```diff
@@ -44,3 +43,40 @@ tentative_process <- function(.data, fx, param){
...
   out
 }
+
+
+#' Checks required columns exsist
+#'
+#' @param .data Data to apply the tfrmt to
+#' @param tfrmt tfrmt object to apply to the data
+#'
+#' @return formatted tibble
```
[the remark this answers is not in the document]
statasaurus: True

statasaurus, line 53: `#' @return Error if columns don't exist otherwise nothing`

PR #57, opened 2022-04-25
Sci notation

karima662:
I have added the `scientific` parameter to the frmt() function and added some code to the apply_frmt() function which enables the user to define a string (of scientific notation) to append to values.


tests/testthat/test-apply_tfrmt.R
```diff
@@ -20,6 +20,31 @@ test_that("applying frmt", {
...
+  )
+
+  expect_equal(
+    sample_df_frmted$x,
+    c("1234.6 x10^3",
+      "345.7  x10^2",
+      " 56.8   x10^1",
+      "4567.9 x10^3",
+      "  8.9    x10^0")
+  )
```
thebioengineer, line 44: Thanks for adding the test! This is super helpful.

When we refer to scientific notation, the coefficient (the `A` in `A x 10^B`) is always a number between 1 and 10. Right now the code is doing the rounding as described in the format, then applying the scientific notation to get the coefficient. 

So we would expect to have the values:
```r
c("  1.2 x10^3",
  "  3.5 x10^2",
  "  5.7 x10^1",
  "  4.6 x10^3",
  "  8.9 x10^0")
```

I think the changes to the code are that you'll want to apply the formatting to scientific at the beginning of `apply_frmt.frmt` (pre-rounding), then going from there. Let us know if that doesn't make sense/you have questions!

karima662: I've managed to get the expected numeric values printed, but there still seems to be some issue around the padding (spaces/indentation - not sure what it's supposed to be called 😅) which affects the other tests.

PR #65, opened 2022-04-28
Address #64

### LINKED ISSUE #64: expr_to_filter.quosures does not handle named list containing ".default" value
``` 
expr_to_filter.quosures(cols = vars(grp1, grp2), val = list(grp1 = ".default", grp2 = "val")) 
``` 
returns 
```
grp1 %in% c('.default')&grp2 %in% c('val')
```

thebioengineer:
Resolve #64 
Add tests for expr_to_filter to confirm behavior in a few different scenarios
Resulted in some changes to error messages to _hopefully_ make them easier to understand? Open for feedback & comment.

### DIFF
--- a/R/apply_table_frmt_plan.R
+++ b/R/apply_table_frmt_plan.R
@@ -117,7 +117,9 @@ expr_to_filter <- function(cols, val){
 }
 
 expr_to_filter.quosure <- function(cols, val){
-  if(all(val == ".default")){ # This is all so it works when there is a list
+
+  # This is all so it works when there is a list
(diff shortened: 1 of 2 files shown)


thebioengineer:
> 

This is so weird... all converts `logical(0)` to TRUE....

thebioengineer:
Made an update to handle the case you identified, @bzkrouse.
Since we are leaning towards requiring explicitness rather than implicitness, I agree that we should have to write .default not just pass TRUE when a group doesnt have a val defined

statasaurus (approved):
💯

PR #68, opened 2022-05-06
Add to mock table functionality

bzkrouse:
- If >1 param per unique label value, force unnest() after pivot_wider() and print console message notifying user
- build on make_mock_data() so it can handle multiple grouping variables and/or group_val as nested list


statasaurus (approved):
This looks really great! I like how it looks

PR #69, opened 2022-05-06
accept tidyselect in element_align

### LINKED ISSUE #56: `col_align` doesn't work with tidyselect variable
It doesn't throw an error but it also doesn't seem to work 
image
image

bzkrouse:
resolves #56


statasaurus (approved):
Looks great!

PR #74, opened 2022-05-16
Sci Notation ellis

thebioengineer:
Branched off #57, added more tests and simplified the logic.


statasaurus (approved):
🐅 👍 (think tony the tiger)

PR #77, opened 2022-05-18
Spanning grp

statasaurus:


R/apply_row_grp_plan.R
```diff
@@ -0,0 +1,210 @@
...
+      map_dfr(function(lone_dat){
+        new_row <- lone_dat %>%
+          select(!!!top_grouping, !!label) %>%
+          mutate(!!label := !!last(group)) %>%
+          distinct()
+
+        lone_dat %>%
+          mutate(!!label := str_c(indent, !!label)) %>%
+          bind_rows(new_row, .) %>%
+          mutate(across(-(!!!group), ~replace_na(., "")))
```
bzkrouse, line 203:
```suggestion
          mutate(across(-c(!!!group), ~replace_na(., "")))
```

R/print.R
```diff
@@ -33,16 +29,47 @@ print_mock_gt <- function(tfrmt, .data = NULL, .default = 1:3, n_cols = 3) {
...
+
+#' Do all the formatting for the GT
+#'
+#' @param .data cleaned dataset
+#' @param tfrmt tfrmt
+#'
+#' @return GT object
+#' @noRd
+cleaned_data_to_gt <- function(.data, tfrmt){
+  if(is.null(tfrmt$row_grp_style) && length(tfrmt$group) > 0){
```
bzkrouse, line 50: Just pointing out that if someone passes an empty row_group_plan() then we apply the defaults: indenting lower-level groups & output data is grouped by highest level grouping, while here if no row_group_plan() then we do no indentation and group by all vars, is that right?

bzkrouse (approved):
Looking great and working great!!! ⭐ 🎉 
Just 1-2 edits and a comment for you!

R/apply_row_grp_plan.R
```diff
@@ -0,0 +1,210 @@
...
+  if(is.null(element_row_grp_loc)){
+    indent = "  "
+  } else if(element_row_grp_loc$location %in% c("spanning", "column") & length(group) > 0){
+    group = group[-1]
+    indent = element_row_grp_loc$indent
+  } else {
+    indent = element_row_grp_loc$indent
+  }
+
+  while(length(group) > 0 & !is.null(label)){
```
bzkrouse, line 188: this is nice!!!

R/theme_element.R
```diff
@@ -1,23 +1,53 @@
 
-element_block <- function(background_fill,
-                          outline,
-                          post_space = c(NULL, " ", "", "------")){
+#' Element Row Group Location
+#'
+#' @param location Location of the grouping. Where 'indented' has all groups in
```
bzkrouse, line 4: #' @param location Location of the row group labels. Specifying 'indented' combines all group and label variables into a single column with each sub-group indented under its parent. 'spanning' and 'column'
#'  retain the highest level group variable in its own column and combine all remaining group and label variables into a single column with sub-groups indented. The highest level group column will either be printed as a spanning header or in its own column in the gt.
#' @param indent A string of the number of spaces you want to indent

R/print.R
statasaurus: Yeah I put that comment in somewhere. I am not sure if I love that behavior but I was going to try and play around a bit and see how it shakes out. Unless you think we should just change it now. I wanted the indentation to be the default cause that feels like the most common approach but I am might be wrong

PR #80, opened 2022-05-19
update to using col_plan

thebioengineer:
various changes to span_plan to allow col_plan ideas, updating tests.

Some tests were failing locally, but unsure if the failures will continue. 

LMK if you have questions!


statasaurus:
I am going to pick this up. It is mostly done I am just going to pull in main and finish off

tests/testthat/test-col_plan.R
```diff
@@ -0,0 +1,281 @@
...
+})
+
+test_that("From the spanning structure get cols to span across", {
+
+  s1 <- span_structure(
+    label = "Test Label",
+    vars(A,B)
+  )
+
+  s1 <- span_structure(
```
bzkrouse, line 67: duplicate

tests/testthat/test-col_plan.R
```diff
@@ -0,0 +1,281 @@
...
+      `__tlang_span_structure_column__1` = c(NA, NA, "test label", rep(NA, 7)),
+      columns = paste0("col",1:10),
+      group = "groupvar",
+      label = "labels",
+      param = "params",
+    )
+  )
+
+  nested_spans_wide_data <- apply_span_structures_to_data(tfrmt_obj = tfrmt_obj_nested_spans, x = input_data)
+  print_mock_gt(tfrmt_obj_nested_spans, input_data)
```
bzkrouse, line 153:
```suggestion
```

tests/testthat/test-col_plan.R
```diff
@@ -0,0 +1,281 @@
...
+  expect_error(
+    col_plan(
+      func("invalid entry")
+    ),
+    paste0(
+      "Invalid entry: `func(\"invalid entry\")`\n",
+      "Only span_structures (`span_structure()`),",
+      " selection helpers (See <https://tidyselect.r-lib.org/reference>),",
+      "  or unquoted expressions representing variable names  can be entered",
+      " as contents. Changes the names of individual variables using",
```
bzkrouse, line 252:
```suggestion
      " as contents. Changing the names of individual variables using",
```

tests/testthat/test-col_plan.R
```diff
@@ -0,0 +1,281 @@
...
+    span_structure(
+      label = "Test Sub Label",
+      vars(C,A)
+    ),
+    vars(B)
+  )
+
+  sample_df <- data.frame(A = character(), B = character(), C = character(), D = character())
+
+  s1_cols <- span_col_select(s1, data = sample_df)
```
bzkrouse, line 93: can we add a test representing column removal as well?

R/utils_tfrmt.R
```diff
@@ -23,26 +28,49 @@ apply_tfrmt <- function(.data, tfrmt, mock = FALSE){
...
     column = tfrmt$column,
     mock = mock
-  )
+  ) %>%
+    tentative_process(apply_col_align_plan, tfrmt$col_align,
+                      tfrmt$column, tfrmt$values,
+                      fail_desc= "Unable to align dataset")
+
+  ## append span structures to dataset for handling post-this function
+  if(!is.null(tfrmt$col_plan$span_structures)){
```
bzkrouse, line 37: any reason why we don't use tentative_process here?

tests/testthat/test-make_mock_data.R
```diff
@@ -199,3 +199,36 @@ test_that("Check mock when value it missing", {
...
 
 
+test_that("Test when no body_style is present", {
+  tfrmt_obj_one_span <- tfrmt(
+    label = label,
+    group = group,
+    param = param,
+    column = vars(columns),
+    col_plan = col_plan(
+      group,label,
```
bzkrouse, line 209: if these don't get specified here, do they end up last?

R/col_plan.R
```diff
@@ -0,0 +1,497 @@
...
+    }
+  }
+  as.character(as.list(x)[[1]]) %in% c("vars","quo")
+}
+
+check_col_plan_dots <- check_span_structure_dots
+
+
+## ---------------------------------------
+## apply col_plan
```
bzkrouse, line 165: delete?

R/col_plan.R
```diff
@@ -0,0 +1,497 @@
...
+      "as well as span_structures in `col_plan`.\n",
+      "The use of only one approach is permitted. ",
+      "Select a single column or remove span_structures from `col_plan()`"
+    )
+  }
+
+}
+
+
+### amend the tfrmt column argument to include new spanning columns
```
bzkrouse, line 306: delete?

R/col_plan.R
```diff
@@ -0,0 +1,497 @@
...
+get_span_structure_dots.span_structures <- function(x){
+  do.call('c',lapply(x$span_cols, get_span_structure_dots))
+}
+
+
+## -----------------------------------------------
+## check in tfrmt that the column and col_plan
+## are compatable
+##-----------------------------------------------
+check_column_and_col_plan <- function(x){
```
bzkrouse, line 286: is this being used?

R/col_plan.R
```diff
@@ -0,0 +1,497 @@
+#' Define the Column Plan
```
bzkrouse, line 1: This is very thorough but is currently moreso than other "plans" in the package. I would love for them to be more consistent in terms of checking, classes, etc. But probably something that can be tabled for later!

bzkrouse (approved):

R/utils_tfrmt.R
statasaurus: Because it takes in a whole tfrmt so would try to work on any tfrmt rather than just if there is a span_structure

tests/testthat/test-make_mock_data.R
statasaurus: well technically last in the dataset. But because of how gt works it will automatically pull the labels to the left. I don't think you could have grouping and labels on the right hand side

R/utils_tfrmt.R
```diff
@@ -23,26 +28,49 @@ apply_tfrmt <- function(.data, tfrmt, mock = FALSE){
...
   }
 
-  tbl_dat_wide %>%
-    tentative_process(arrange, tfrmt$sorting_cols) %>%
-    tentative_fx(apply_row_grp_plan, tfrmt$row_grp_style, tfrmt$group, tfrmt$label) %>%
-    tentative_process(select, tfrmt$col_select)%>%
-    tentative_fx(apply_col_align_plan, tfrmt$col_align)
+  tbl_dat_wide <- tbl_dat_wide %>%
+    tentative_process(arrange_enquo, tfrmt$sorting_cols, fail_desc= "Unable to arrange dataset") %>%
+    #Select before grouping to not have to deal with if it indents or not
```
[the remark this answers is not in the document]
bzkrouse: on second thought, the cdisc pilot demog table is an example where we need a second grouping variable (defining whether the row groups are categorical or continuous) so we can put a post space after each. If the goal is to allow people to drop group variables, which would be useful for that table, we'll need a different workflow.

tests/testthat/test-col_plan.R
statasaurus: I have added this, but we probably need to update these tests anyway cause the

R/utils_tfrmt.R
```diff
@@ -23,26 +28,49 @@ apply_tfrmt <- function(.data, tfrmt, mock = FALSE){
...
+      tfrmt,
+      tbl_dat
+    )
+  }else{
+    tbl_dat_span_cols <- tbl_dat
+  }
+
+  tbl_dat_wide <- quietly(pivot_wider)(
+    tbl_dat_span_cols,
+    names_from = c(starts_with(.tlang_struct_col_prefix), !!!tfrmt$column),
```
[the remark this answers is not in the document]
statasaurus: In tests yes but I don't think it makes sense to always check this behaviour every time we run the code

R/utils_tfrmt.R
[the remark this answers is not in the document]
statasaurus: You can't drop grouping variables. They either get merged together like with multiple grouping variables or merged into the labels. But they can never be fully dropped

R/col_plan.R
statasaurus: yes in the tfrmt function

PR #90, opened 2022-05-23
Update col align tests

bzkrouse:


R/col_align.R
```diff
@@ -94,15 +97,17 @@ apply_col_align_plan <- function(.data, align_plan, columns, val_col){
 
   .data %>%
     left_join(align_spec, by = as_label(last_col)) %>%
-    group_by(!!!columns) %>%
+    mutate(across(c(!!!column), fct_inorder)) %>%
```
bzkrouse, line 100: @statasaurus added this to keep the order of the columns as they appear in the data. otherwise it will order alphabetically

PR #92, opened 2022-05-26
Col plan allow grp rmv

bzkrouse:
I went ahead and implemented some updates that would allow the user to drop unneeded group columns from the final table. For example, if there are extra group variables needed for grp blocks, like in the CDISC demog table. I wasn't sure if we were expecting this to work in the current branch or not. Lmk your thoughts!


R/apply_row_grp_plan.R
```diff
@@ -12,14 +12,8 @@
 #' @importFrom tidyselect everything
 #' @importFrom rlang !!!
 apply_row_grp_plan <- function(.data, row_grp_plan, group, label = NULL, ...){
```
bzkrouse, line 14: probably makes sense to rename this if we are going with these updates

bzkrouse:
@statasaurus did the updates we talked about:

- col_plan behaves like dplyr::select
- location = "noprint" option in row_grp_plan that is actually passed into the col select for col_plan

Also drafted readme, building blocks, body plan, and row group plan vignettes. 
Made empty files as placeholders for the rest. 

Any questions let me know!

statasaurus:
I know this still fails on a mac. Not sure why. I am just going to merge this into main so I can update names etc.

PR #102, opened 2022-05-31
Update Actions

thebioengineer:
Actions are failing for mac. Updated to reflect latest action set from usethis crossed with what we were already doing.


statasaurus (approved):
Looks great thanks for sorting this

PR #107, opened 2022-06-01
Col plan select

thebioengineer:
adding ability to col_plan to set new names in dataset outputs


tests/testthat/test-col_plan.R
```diff
@@ -186,6 +213,19 @@ test_that("From col plan spanning structures, get df to add to data",{
     )
   )
 
+  one_span_wide_data_rename <- apply_span_structures_to_data(tfrmt_obj = tfrmt_obj_one_span_rename, x = input_data)
+
+  expect_equal(
+    one_span_wide_data_rename,
+    tibble(
+      `__tlang_span_structure_column__1` = c(NA, NA, "test label", rep(NA, 7)),
+      columns = paste0("col",1:10),
```
bzkrouse, line 222: is col3 supposed to be renamed here?

thebioengineer: Renaming of values does not occur until select_col_plan. Here we are just trying to make the data format the same between span_strcture and multiple column formats

R/zzz.R
```diff
@@ -1,4 +1,4 @@
-globalVariables(c(".", ".data"))
+globalVariables(c(".", ".data", ".rename_col", ".original_col","df_names","new_name_in_df"))
```
bzkrouse, line 1: @thebioengineer Can you explain to me why these are necessary to put here?

thebioengineer: R CMD CHECK was complaining about no global reference to these. It is due to the tidyverse code I am using in select_col_plan.

PR #109, opened 2022-06-02
address issues with renaming cols with multi-column names

thebioengineer:
Found issue with renaming columns with multiple columns. Added some robustness and tests


statasaurus (approved):
🦄

Issue #93, opened 2022-05-27
allow user to pass bare/string into sorting_cols


bzkrouse:
currently only accepts quosure(s)

Issue #99, opened 2022-05-31
Warning when passing a qou() to column rather than vars 


statasaurus:
```{r}
── Warning (test-apply_frmt.R:401:3): mocks return correctly ───────────────────
Unquoting language objects with `!!!` is deprecated as of rlang 0.4.0.
Please use `!!` instead.

  # Bad:
  dplyr::select(data, !!!enquo(x))

  # Good:
  dplyr::select(data, !!enquo(x))    # Unquote single quosure
  dplyr::select(data, !!!enquos(x))  # Splice list of quosures

This warning is displayed once per session.
Backtrace:
  1. ... %>% pull(x)
       at test-apply_frmt.R:401:2
 32. rlang:::warn_deprecated(x, id = y)
 33. base .Signal(msg = msg)
````

PR #110, opened 2022-06-02
sorting_cols as a var and convert quo to vars

thebioengineer:
- Update ability to convert quo to vars
- Add sorting_cols to list of args to convert to vars
- update vignettes to use c instead of vars for column and sorting_cols


thebioengineer:
resolves #93 and #99

statasaurus (approved):
Looks great thanks for the update

PR #114, opened 2022-06-06
col_plan vignette

bzkrouse:


statasaurus (approved):

vignettes/col_plan.Rmd
```diff
@@ -14,6 +14,283 @@ knitr::opts_chunk$set(
...
+    "g2", "rowlabel3",  "col1"  ,  "value",    3,
+    "g2", "rowlabel3",  "col2"  ,  "value",    3,
+    "g2", "rowlabel3",  "mycol3",  "value",    3,
+    "g2", "rowlabel3",  "col4"  ,  "value",    3,
+    "g2", "rowlabel3",  "mycol5",  "value",    3)
+  
+```
+
+
+## No column spanners
```
thebioengineer, line 62: I think a title that is more descriptive of what it is, rather than what it isnt may be more helpful for folks.
```suggestion
## Simple Column Selection
```

vignettes/col_plan.Rmd
```diff
@@ -14,6 +14,283 @@ knitr::opts_chunk$set(
...
+    param = parm,
+    values = val,
+    column = my_col,
+    body_plan = body_plan(
+      frmt_structure(group_val = ".default", label_val = ".default", frmt("x"))
+    ),
+    col_plan = col_plan(
+      group, 
+      label, 
+      starts_with("col"),
```
thebioengineer, line 99: Should we default to keeping these values when they are defined? I am feeling as though the users shouldn't need to define them. what are everyone's thoughts?

vignettes/col_plan.Rmd
```diff
@@ -14,6 +14,283 @@ knitr::opts_chunk$set(
...
+    )
+  ) %>% 
+  print_to_gt(dat)
+```
+
+
+
+## Column spanners
+
+ The `col_plan` accepts a combination of column names and/or span structure (`span_structure`) objects. Recall that `*_structure` objects are for modifying the *contents* (i.e., text) of the table; in the `col_plan`, we use `span_structure`s to add column spanners. These are placed strategically in the `col_plan` to inform `tlang` of the hierarchy of spanners vs. columns.
```
thebioengineer, line 111:
```suggestion
 The `col_plan` accepts a combination of tidyselect expressions, column names, and/or span structure (`span_structure`) objects. Recall that `*_structure` objects are for modifying the *contents* (i.e., text) of the table; in the `col_plan`, we use `span_structure`s to add column spanners. These are placed strategically in the `col_plan` to inform `tlang` of the hierarchy of spanners and columns.
 
A `span_structure` is constructed from a label which defines the spanner text, and a combination of tidyselect expressions and column names.
```

vignettes/col_plan.Rmd
```diff
@@ -14,6 +14,283 @@ knitr::opts_chunk$set(
...
+        starts_with("col")
+        ),
+      new_col_3 = mycol3,  
+     - mycol5            
+    )
+  ) %>% 
+  print_to_gt(dat)
+```
+
+We can also do any number of hierarchical spanners:
```
thebioengineer, line 139:
```suggestion
A `span_structure` can also contain one or more child `span_structure` entries, defining a hierarchy of spanners:
```

vignettes/col_plan.Rmd
```diff
@@ -14,6 +14,283 @@ knitr::opts_chunk$set(
...
+      starts_with("col"),
+      new_col_3 = mycol3,  
+      -mycol5            
+    )
+  ) %>% 
+  print_to_gt(dat)
+```
+
+
+When there are multiple `column` variables provided, `tlang` concatenates them behind the scenes using a special delimiter. Therefore, the underlying column names in the final table are a combination of the spanner(s) and the lowest level column label. In order to change the order of spanning groups, we may use`contains()` to reference any portion of those names.
```
thebioengineer, line 252: Not sure if we want to share the details on how `tlang` behaves, but understand why to make it so folks understand how they can use `contains()`. 

```suggestion
When there are multiple `column` variables provided, `tlang` concatenates them behind the scenes using a special delimiter. Therefore, the underlying column names in the final table are a combination of the spanner(s) and the lowest level column label. In order to change the order of spanning groups, we may use `contains()` to reference any portion of those names. 
```

vignettes/col_plan.Rmd
```diff
@@ -14,6 +14,283 @@ knitr::opts_chunk$set(
...
+      -mycol5            
+    )
+  ) %>% 
+  print_to_gt(dat)
+```
+
+
+When there are multiple `column` variables provided, `tlang` concatenates them behind the scenes using a special delimiter. Therefore, the underlying column names in the final table are a combination of the spanner(s) and the lowest level column label. In order to change the order of spanning groups, we may use`contains()` to reference any portion of those names. 
+
+Reordering with multiple `column` variables is simplest when the lowest level column variable contains unique values (i.e., there is 1 value per column in the final table), like in the example above. Let's suppose we want to move the "col 4" column to the beginning and also reorder "col1" and "col2". To achieve this, we can just select "col 4" and then "col2" and finally all the other columns.
```
thebioengineer, line 254:
```suggestion
Reordering with multiple `column` variables is simplest when the lowest level column variable contains unique values (i.e., there is 1 value per column in the final table), like in the example above. Let's suppose we want to move the "col 4" column to the beginning and also reorder "col1" and "col2". To achieve this, we can select "col 4" and then "col2" and finally all the other columns. 
```

thebioengineer (changes requested):

vignettes/col_plan.Rmd
bzkrouse: I'm going to revise so it's somewhere in the middle.

vignettes/col_plan.Rmd
bzkrouse: I tend to agree. I like that people have control over order/dropping them, but i think it would be a nice default to keep them in if not specified.

thebioengineer (approved):
thanks for the changes! 🥇

PR #116, opened 2022-06-07
Adding a Vignette for unusual tables  

statasaurus:


vignettes/unusual_tables.Rmd
```diff
@@ -0,0 +1,85 @@
...
+library(tidyr)
+```
+
+
+# Multiple columns of stub
+
+It is not all that unusual for listing (and some tables) to have multiple stub columns. When this happens it is often easier to not use thee stub functions for gt. An example of a table like this is the "Summary of Number of Subjects by Site" from the CDISC pilot.
+![Image of the Summary of Number of Subjects By Site table from the CDISC pilot](images/summary_subj_by_site.png)
+
+To make this table the values will be long with "Pooled Id" and "Site Id" as separate columns like normal. We also will still need a column for the parameters even though they are all the same.
```
bzkrouse, line 30:
```suggestion
To make this table the values will be long with "Pooled Id" and "Site Id" in their own columns, as if they were group or label variables. We also will need a column for the parameters even though they are all the same. 
```

vignettes/unusual_tables.Rmd
```diff
@@ -0,0 +1,85 @@
...
+                   "Xanomeline Low Dose (N=84)",
+                   "Xanomeline High Dose (N=84)",
+                   "Total (N=254)"),
+           col2 = c("ITT", "Eff", "Com")) %>%
+  mutate(val = rpois(216, 15), # Here I am just faking the data for display purposes 
+         param = "val")
+
+```
+
+Once we have the data in the standard ARD format we can make the tfrmt. What makes this tfrmt different is we won't include group or label. This will make it so the table will just be ordered based on the `col_plan()`. Because the `col_plan()` follows the conventions of `select()` we are going to start with the lowest level columns, and then select the order of the spanninig columns.
```
bzkrouse, line 63:
```suggestion
Once we have the data in the standard ARD format we can make the tfrmt. What makes this tfrmt different is we won't include group or label. This will make it so the table will just be ordered based on the `col_plan()`. Because the `col_plan()` follows the conventions of `select()` we are going to start with the lowest level columns, and then select the order of the spanning columns. 
```

bzkrouse (changes requested):
Looks great! I added a couple of edits (take them or leave them!) and a suggestion to avoid the terminology of stub  unless we plan to adopt it throughout

vignettes/unusual_tables.Rmd
```diff
@@ -0,0 +1,85 @@
...
+
+```{r setup}
+library(tlang)
+library(tibble)
+library(dplyr)
+library(tidyr)
+```
+
+
+# Multiple columns of stub
```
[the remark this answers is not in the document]
statasaurus: I updated this

Issue #113, opened 2022-06-06
`col_plan` throws error with varying levels of `span_structure` nesting


bzkrouse:
Error: unable to subset data columns

```
dat <- tribble(
    ~group,     ~label,  ~my_col,    ~parm, ~val,
    "g1", "rowlabel1",  "col1",  "value",    1,
    "g1", "rowlabel1",  "col2",  "value",    1,
    "g1", "rowlabel1",  "mycol3",  "value",    1,
    "g1", "rowlabel1",  "col4",  "value",    1,
    "g1", "rowlabel1",  "mycol5",  "value",    1,
    "g1", "rowlabel2",  "col1",  "value",    2,
    "g1", "rowlabel2",  "col2",  "value",    2,
    "g1", "rowlabel2",  "mycol3",  "value",    2,
    "g1", "rowlabel2",  "col4",  "value",    2,
    "g1", "rowlabel2",  "mycol5",  "value",    2,
    "g2", "rowlabel3",  "col1",  "value",    3,
    "g2", "rowlabel3",  "col2",  "value",    3,
    "g2", "rowlabel3",  "mycol3",  "value",    3,
    "g2", "rowlabel3",  "col4",  "value",    3,
    "g2", "rowlabel3",  "mycol5",  "value",    3)
```

works:
```tfrmt(
    group = group,
    label = label,
    param = parm,
    values = val,
    column = my_col,
    col_plan = col_plan(
      group, 
      label, 
      span_structure(
        "column cols",
        starts_with("col")
        ),
      span_structure(
        "my cols",
        new_col_3 = mycol3,
        mycol5
    ))
  ) %>% 
  print_to_gt(dat)
```

throws error:
```
tfrmt(
    group = group,
    label = label,
    param = parm,
    values = val,
    column = my_col,
    col_plan = col_plan(
      group, 
      label, 
      span_structure(
        "column cols",
        span_structure(
          "cols 1,2",
          col1, col2
        ),
          col4
        ),
      span_structure(
        "my cols",
        new_col_3 = mycol3,
        mycol5
    ))
  ) %>% 
  print_to_gt(dat)
```

thebioengineer:
Mostly sorted this out now. Having difficulties getting '-' to propagate now though. Ie col_plan(-col_4) with multiple cols

thebioengineer:
correction already merged. additional edits in #135

PR #117, opened 2022-06-07
Col plan multilayer select

thebioengineer:
addresses #113 , and adds additional behavior to handle `-` in col_plans when the field is renamed.

Adds tests for these cases as well.

It was found that with _multiple_ levels of spanning headers, there were differences in how the names were created for matching in select_col_plan and clean_spanning_col_names. these are now in line.


thebioengineer:
tbh not sure how to correct the failing runs. I'll look into how I can get those environments and test in them

R/col_plan.R
```diff
@@ -343,27 +355,47 @@ select_col_plan <- function(data, tfrmt){
...
-          new_name_in_df = case_when(
-            new_name_in_df == .original_col ~ NA_character_,
-            TRUE ~ new_name_in_df
-          )
-        )
+          new_name_in_df = remove_empty_layers(.data$new_name_in_df, n_layers),
+          new_name_in_df_output = remove_empty_layers(.data$new_name_in_df_output, n_layers)
+        ) %>%
+        mutate(
+          new_name_quo = map2(.data$new_name_in_df, .data$.removal_identifier_col, function(x, y) {
```
statasaurus, line 365: Given this already an anonymous function can you make it a named function and pull it out this easier to maintain

R/col_plan.R
```diff
@@ -343,27 +355,47 @@ select_col_plan <- function(data, tfrmt){
...
+            x_text <- tryCatch({
+              x_lang <- parse(text = x)[[1]]
+              if (is_valid_tidyselect_call(x_lang)) {
+                x
+              } else{
+                paste0("`", x, "`")
```
statasaurus, line 371: Can you add additional comments here?

statasaurus (approved):
My comments here are more take it or leave it.

tests/testthat/test-col_plan.R
```diff
@@ -457,6 +457,161 @@ test_that("Build simple tfrmt with multiple columns and with renaming duplicated
...
+      col1,
+      col2,
+      col4,
+      new_col_3 = mycol3,
+      mycol5
+    )
+  )
+
+
+  suppressMessages({
```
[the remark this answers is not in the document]
thebioengineer: TBH I don't recall why I was doing the warning suppression here. We don't want the messages b/c its the message which rows don't get processed by the body plan. We may want to rethink that a bit if there was no body_plan provided whether to return that message.

R/apply_tfrmt.R
```diff
@@ -71,12 +71,17 @@ tentative_process <- function(.data, fx, ..., fail_desc = NULL){
...
       safely(fx)(...)
     if(!is.null(out[["error"]])){
-      out <- .data
       if(is.null(fail_desc)){
-        message("Unable to to apply formatting", format(substitute(fx)))
-      }else{
-        message(fail_desc)
+        fail_desc <- paste0("Unable to to apply ",format(substitute(fx)),".")
       }
+      fail_desc <-paste0(
```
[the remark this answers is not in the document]
thebioengineer: NP. I was using it to debug too. So figured we should keep that.

R/col_plan.R
thebioengineer: NP. will do.

PR #118, opened 2022-06-07
Efficacy tbl

bzkrouse:
- add efficacy data/example (@statasaurus we can move this to the unusual tables vignette later if you'd like)
- ensure apply_frmt keeps decimal places after rounding
- fill new NAs with "" during the pivot_wider


vignettes/examples.Rmd
```diff
@@ -526,3 +526,297 @@ tfrmt(
...
+  values = value 
+) %>%
+  print_to_gt(efficacy_data) %>% 
+  tab_options(
+    container.width = 800
+  ) 
+  
+```
+
+Through judicious use of `body_plan`'s frmt`, `frmt_combine`, and `frmt_when`, we can conditionally format each of the different pieces of results. For the summary statistics, we have the number of observations (n), the mean and standard deviation (mean, sd), and the median and range (median, min, max). For the models, we have the p-value (p.value) and the least squares mean difference (diff) as well as its associated standard error (diff_se) and 95% confidence interval (diff_lcl, diff_ucl).
```
thebioengineer, line 604:
```suggestion
Through judicious use of `body_plan`'s `frmt`, `frmt_combine`, and `frmt_when`, we can conditionally format each of the different pieces of results. For the summary statistics, we have the number of observations (n), the mean and standard deviation (mean, sd), and the median and range (median, min, max). For the models, we have the p-value (p.value) and the least squares mean difference (diff) as well as its associated standard error (diff_se) and 95% confidence interval (diff_lcl, diff_ucl). 
```

statasaurus (approved):
🐘 Amazing! This looks great 😸

Issue #60, opened 2022-04-26
Print function for frmt_when


statasaurus:

karima662:
Test with the following:

combine <- frmt_combine(expression = "{Min}, {Max}",
                        Min = frmt("XX"),
                        Max = frmt("XXX"))

when <- frmt_when("==100"~ frmt(""),
                  "==0"~ "",
                  "TRUE" ~ frmt("(XXX.X%)"))

combine
when

karima662:
I think this issue can be closed now

PR #121, opened 2022-06-08
Print frmt when

karima662:
See issue #60 for more info

### DIFF
--- a/R/frmt_utils.R
+++ b/R/frmt_utils.R
@@ -69,12 +69,15 @@ print.frmt_combine <- function(x,...){
 
 #' @export
 format.frmt_when <- function(x, ...){
+  lhs <- map_chr(x, f_lhs)
+  rhs <- map(x, f_rhs) %>% map_chr(format)
+
   frmt_str <- cat(
-    "< frmt_when | ", "\n",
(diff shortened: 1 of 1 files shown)


R/frmt_utils.R
```diff
@@ -67,6 +67,24 @@ print.frmt_combine <- function(x,...){
   cat(format(x),sep = "\n")
 }
 
+
+#' @export
+format.frmt_when <- function(x, ...){
+  frmt_str <- cat(
+    "< frmt_when | ", "\n",
+    "LHS: ", paste0(map(x, rlang::f_lhs) %>% map(format) %>% unlist(), collapse = ", "), "\n",
```
statasaurus, line 75:
```suggestion
    "LHS: ", paste0(map(x, f_lhs) %>% map(format) %>% unlist(), collapse = ", "), "\n",
```
Can you remove the double colons. If this breaks the code it means you need to import these functions you can so that by adding 
```r
#' @importFrom rlang f_lhs rf_rhs 
```

statasaurus (changes requested):
Looks really good. Other than removing the double colon I think we are good to go 😄

PR #132, opened 2022-06-14
Labs and sigdig

### LINKED ISSUE #127: function to build `body_plan` according to significant digits spec
Defining significant digits for different variable values is already a standard part of specs. 
This enhancement will provide an option in `tfrmt()` to:
 - Pass this spec as a dataset
 - Optionally modify a set of default `param` values and their added precision. Something like: ```list(0 = c("min", "median", "max"), 1 = c("mean", "geomean"), 2 = c("sd","cv")```, where the numeric value represents what to add on top of the significant digit specified in the specs. 
 
The dataset in conjunction with the param list will be passed to a function which will build the relevant pieces of the `body_plan`

bzkrouse:
- resolves #127 
- add labs example data

example of use:

```
dat_sigdig <- dplyr::tribble(
  ~group1,  ~ group2, ~ sigdig,
  ".default",  ".default", 0,
  "CHEMISTRY",  "ALANINE AMINOTRANSFERASE", 0,
  "CHEMISTRY", "BILIRUBIN", 0,
  "CHEMISTRY", "CHOLESTEROL", 1,
  "CHEMISTRY", "EOSINOPHILS", 1
  
) 
labs <- labs_data %>% 
  select(-c(value, starts_with("ord"))) %>% 
  filter(group1 %in% unique(dat_sigdig$group1),
         group2 %in% unique(dat_sigdig$group2)) %>% 
  unite("cols", c(col1, col2))


  tfrmt_builder(data = dat_sigdig, 
              group = c(group1, group2),
              param_defaults  = param_set(),
              missing = NULL) %>%
   layer_tfrmt(
     tfrmt(
       column = cols,
       param = param,
       label = rowlbl)) %>% 
  print_mock_gt(., .data = labs)

```


bzkrouse:
@statasaurus @thebioengineer Holding off on merging in main for now due to other issues.

statasaurus (approved):
I have just made some cosmetic changes, 
* I updated the wording of the message to be more specific that the tfrmt would need [OOO] in order to be used with data when trying to print_to_gt
* I moved the builder functions to frmt_utils and body_plan as they aren't really specific to the sigdigs and we might want to used the in the future 
* added code to sort the defaults, so if something is .default all the way across it is higher order than just one group .default and all .defaults are before more specific groups 
* I added an example to tfrmt_sigdig 

also you didn't  have tfrmt_sigdig exported, was there a reason for that?

PR #137, opened 2022-06-17
suggestions to avoid using text parsing

thebioengineer:
my suggestions to construct body plan/ format structure/ frmts w/o using string pasting. 

Don't need to take it, just a suggestion


bzkrouse (approved):
⭐ ⭐ ⭐ ⭐ ⭐

PR #138, opened 2022-06-17
Resolve failing main & allow tfrmt to be made in functions

thebioengineer:
Addresses the failing main tests. Reason was when layering tfrmts with un-defined vars or quos (ie one tfrmt has sorting_cols defined, one does not), results in trying to compare null values, which errors. This has now been corrected.

Also corrected environments that the function is looking for values in so now tfrmts can be constructed in other functions as expected. 

Finally, corrected warning when a frmt_structure was built with a frmt_combine and when printed complained about multiple params.


R/tfrmt.R
```diff
@@ -51,7 +53,7 @@ tfrmt <- function(
   ...
   ){
 
-  tfrmt_el <- tfrmt_find_args(...)
+  tfrmt_el <- tfrmt_find_args(..., env = environment(), parent_env = rlang::caller_env())
```
statasaurus: Remove the double colon

R/tfrmt.R
```diff
@@ -128,8 +131,22 @@ quo_get <- function(args, as_var_args = c(), as_quo_args = c(), envir = parent.f
...
+      if(identical(arg_call, quo())){
+        return(arg_call)
+      }
+
       ## try to safely evaluate arg call
-      arg_call_results <-  safely(eval)(arg_call, envir = envir)
+      arg_call_results_envir <-  safely(eval_tidy)(arg_call, env = envir)
+      arg_call_results_parent_env <-  safely(eval_tidy)(arg_call, env = parent_env)
+
+      # browser()
```
statasaurus: Remove the old browser

statasaurus (approved):
This looks good I made a cleaned up the places I said to remove

PR #140, opened 2022-06-20
Rename of the package, there is still an issue with col_plan

statasaurus:


thebioengineer:
Quick look of the diff is showing a number of things being deleted in the .R files that are not related to the package name change, and I think are what is causing col_plan to fail. I'll take a closer look to see details tomorrow. But AFAIK the changes being made in the PR shouldn't be doing that.

statasaurus:
You were right, I thought I had pull from main but I clearly hadn't. I just deleted this and tried again which I think when much better

PR #152, opened 2022-06-22
Prevent promise evaluation warning on layering

thebioengineer:
skip evaluation of quo if is an expected quo var. this prevents a warning that the promise couldn't be evaluated.

basically circular logic is great because its circular. which throws a warning (and an error, but that was caught by safely())


R/tfrmt.R
```diff
@@ -115,13 +115,15 @@ tfrmt_find_args <- function(..., env = parent.frame(), parent_env = parent.env(e
...
 
-#' @importFrom rlang abort frame_call
+#' @importFrom rlang abort frame_call is_quosure
 #' @importFrom dplyr vars
 #' @importFrom purrr safely
 quo_get <- function(args, as_var_args = c(), as_quo_args = c(), envir = parent.frame(), parent_env = parent.env(envir)){
 
   arg_set <- lapply(args, function(arg){
 
+    # browser()
```
bzkrouse, line 125:
```suggestion
```

bzkrouse (approved):
Looks good! ⭐

PR #154, opened 2022-06-22
Sorting cols to mock data

### LINKED ISSUE #148: print_mock_gt() prints error when data is null
Here is a simple example:

```r
mock_tbl <- function(tfrmt_obj){
  tfrmt(
    # Specify columns in the data
    label = label,
    column = column, 
    param = param,
    values = value,
    sorting_cols = c(ord1),
    
    # Specify body plan
    body_plan = body_plan(
      frmt_structure(group_val = ".default", label_val = ".default",
                     frmt_combine(
                       "{count} {percent}",
                       count = frmt("XXX"),
                       percent = frmt_when("==100"~ frmt(""),
                                           "==0"~ "",
                                           "TRUE" ~ frmt("(XX.X%)"))))
    ))
}

mock_tbl() %>%
  print_mock_gt()
```
When no dataset is provided to the `print_mock_gt()` function the default parameters should be used. The function works and the correct table is output as shown: 
image

But an error is also printed in the console:
image

thebioengineer:
Add handling of sorting columns and add tests related. Add `tribble::` to tibble/tribble calls. resolves #148


bzkrouse (changes requested):
This is nice!!! Just need to add map_dfc to the @imports  then good to go!!

Issue #161, opened 2022-06-27
bugs with make mock


statasaurus:
This only makes one line rather than 3 
```{r}
tfrmt(
  param = param,
  label = label, 
  column = column, 
  values = value,
  body_plan = body_plan(
    frmt_structure(group_val = ".default", label_val = ".deafult",
                   frmt_combine("{n} ({percent}%)",
                                n = frmt("xx"),
                                percent = frmt("xxx"))
    )
  )
) %>% 
  print_mock_gt()
```

This produces an error cause label is missing. But, label isn't required when there is data (as long as you only have .defaults) 
```{r}
tfrmt(
  param = param,
  column = column, 
  values = value,
  body_plan = body_plan(
    frmt_structure(group_val = ".default", label_val = ".deafult",
                   frmt_combine("{n} ({percent}%)",
                                n = frmt("xx"),
                                percent = frmt("xxx"))
    )
  )
) %>% 
  print_mock_gt()
```

PR #169, opened 2022-06-28
Mocking data without Lab & providing data with values to mock

thebioengineer:
Resolves issues #161 and not mentioned issue that when data is fed into a mock with a column matching the value argument, it fails.


statasaurus (approved):
Looks great! I just removed the double colon in your code cause we needed to import discard anyway. But other than that it is great

PR #180, opened 2022-06-30
Tfrmt sigdig updates

### LINKED ISSUE #179: tfrmt_sigdig updates
- Accept tfrmt_obj as an argument so it can be layered onto existing tfrmt
- Make group/label optional 
  - Assume that columns in input data are grouping columns if not specified

bzkrouse:
resolves #179 

- make group/label optional
- add ability to pass tfrmt_obj
   - use tfrmt_obj group/label if none provided
- make sure contents & ordering of group param get applied to frmt_structure
- add tests


tests/testthat/test-tfrmt_sigdig.R
```diff
@@ -211,3 +211,111 @@ test_that("tfrmt_sigdig returns a tfrmt", {
...
+  # if only some are specified, assume the rest are groups
+  t_out <- tfrmt_sigdig(dat_sigdig, label = group2)
+
+  expect_equal(t_out$group, vars(group1, lbl), ignore_attr = TRUE)
+  expect_equal(t_out$label, quo(group2), ignore_attr = TRUE)
+
+  # including a group that is not present in the data
+  t_out <- tfrmt_sigdig(dat_sigdig, group = vars(lbl, grp3), label = group2)
+
+  expect_equal(t_out$group, vars(lbl, grp3, group1), ignore_attr = TRUE)
```
statasaurus, line 241: So I am not sure if this would be my expected functionality. I think I would assume if I provide groups it will only use what I provide as groups. I think I would want it to at least warning here saying grp3 isn't in signing. But I would also not have group1 as a group at all . Basically the supplied grouping should supersede the provided data

statasaurus (changes requested):
This looks really good! Thank you for getting it done so quickly. I just have some tweaks about the functionality. If you disagree we can chat more

tests/testthat/test-tfrmt_sigdig.R
bzkrouse: Okay, that makes sense. I think that is clearer  - either you provide groups and it will use that OR it will use the data, but not both. I can update.

statasaurus (approved):
Looks great! Thank you 🦄

PR #181, opened 2022-06-30
Group field rename and check against body plan

thebioengineer:
Per conversation earlier today,
add function to rename group var, and update across body_plan
Also a check in tfrmt to verify the in group and body_plan match.


R/col_plan.R
```diff
@@ -278,28 +278,7 @@ get_span_structure_dots.span_structures <- function(x){
...
-##-----------------------------------------------
-check_column_and_col_plan <- function(x){
-  multi_column_defined <- length(x$column) > 1
-  span_structures_defined <- if(!is.null(x$col_plan)){
-    !is.null(x$col_plan$span_structures)
-  }else{
-    FALSE
-  }
 
-  if(multi_column_defined & span_structures_defined){
```
bzkrouse, line 293: we can have both now? how does it work?

R/tfrmt_checks.R
```diff
@@ -0,0 +1,75 @@
...
+              paste0("  ", format(frmt_struct), collapse = "\n"),
+              "\n"
+              )
+          )
+        }
+      }
+    }
+
+    if(is_invalid_body_plan){
+      abort(is_invalid_body_plan_message, call = caller_call())
```
bzkrouse, line 71: This is really helpful to the user!

R/tfrmt_layer.R
```diff
@@ -121,3 +121,81 @@ layer_tfrmt_arg.body_plan <- function(x, y, ...,  join_body_plans = TRUE){
 }
 
 
+
+#' Remap group values in a tfrmt
```
bzkrouse, line 125: Needs imported functions added to roxygen

bzkrouse (changes requested):
This is such a great addition! Just needs @imports added for update_group. Otherwise ready to go!!

R/tfrmt_checks.R
thebioengineer: Do you think we should add guidance on how to update the groups? I am not sure because this is a check also within the creation of the tfrmt

R/tfrmt_layer.R
thebioengineer: Fixed in c5ceec7

R/tfrmt_checks.R
statasaurus: My feeling is yes. Cause you are clear in the issue. If you make the wording a bit wishywashy like "you might need to update group names using..."

thebioengineer: Added in now! b4e258d

statasaurus (approved):
you are a ⭐

PR #192, opened 2022-07-04
Resolving issue with regex. Now won't out ` around underscores to mat…

statasaurus:
…ch how glue works

### DIFF
--- a/R/frmt_plans.R
+++ b/R/frmt_plans.R
@@ -176,9 +176,9 @@ frmt_combine <- function(expression, ..., missing = NULL){
 
   names(fmt_ls) <- vars_to_fmt
   # Adding ` to expression if not there and there is a space/symbol
-  replace_val <- case_when(
-    str_detect(vars_to_fmt, "^[:alnum:]*$") ~ vars_to_fmt,
-    !str_detect(vars_to_fmt, "^[:alnum:]*$") & !str_detect(vars_to_fmt, "`") ~ paste0("`", vars_to_fmt, "`"),
+  replace_val <-case_when(
+    str_detect(vars_to_fmt, "^[a-zA-Z0-9_]*$") ~ vars_to_fmt,
+    !str_detect(vars_to_fmt, "^[a-zA-Z0-9_]*$") & !str_detect(vars_to_fmt, "`") ~ paste0("`", vars_to_fmt, "`"),
     TRUE ~ vars_to_fmt)
 
   exp_new <- expression


R/apply_frmt_methods.R
```diff
@@ -154,6 +154,9 @@ apply_frmt.frmt_combine <- function(frmt_def, .data, values, mock = FALSE, param
     str_extract_all("(?<=\\{)[^\\}]+(?=\\})") %>%
     unlist()
 
+  # Adding the unquoted version to match while long
+  fmt_param_vals_uq <- str_remove_all(fmt_param_vals, "`")
+
   # Check if unspecified param values are in the dataset
 
   if(!setequal(names(frmt_def$fmt_ls), fmt_param_vals)){
```
bzkrouse, line 162: Should fmt_param_vals be changed to fmt_param_vals_uq?

bzkrouse (commented):
@statasaurus Can you take a look at my comments? The PR looks great otherwise!!

R/apply_frmt_methods.R
```diff
@@ -165,7 +168,7 @@ apply_frmt.frmt_combine <- function(frmt_def, .data, values, mock = FALSE, param
     fmt_to_apply <- frmt_def$fmt_ls[[var]]
 
     .data %>%
-      filter(!!param == var) %>%
+      filter(!!param == str_remove_all(var, "`")) %>%
```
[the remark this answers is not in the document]
thebioengineer: The names of the list expect there to be "`" in the name (line 168). so we either remove the "`" here, or add it in at line 168 conditionally. This is probably simpler code to maintain.

R/apply_frmt_methods.R
```diff
@@ -222,10 +225,10 @@ apply_frmt.frmt_combine <- function(frmt_def, .data, values, mock = FALSE, param
     mutate(
       !!values := case_when(
         .data$.is_all_missing ~ frmt_def$missing,
-        TRUE ~ str_glue(frmt_def$expression) %>% as.character()
+        TRUE ~ str_glue(!!frmt_def$expression) %>% as.character()
```
thebioengineer, line 228: Is there  a reason to `!!` this?

R/apply_frmt_methods.R
statasaurus: We end up needing both. When you are filtering in the data before it goes wide it needs to be without the `, but when you are doing things with the new columns you need `. So there isn't any winning

R/apply_frmt_methods.R
statasaurus: um... for some reason this made it work. I have was having the issue when where "{`val 1} {val2}" would work but 
```{r}
frmt_def$expression <-  "{`val 1} {val2}"
str_glue(frmt_def$expression)
```
wouldn't work. So I just thought I would try the bang bang and it solved it. But not really sure why. (Yes, I realize there is like a 90% chance this will come back to bite me later

PR #194, opened 2022-07-05
add in collapseable code to produce the subset

### LINKED ISSUE #164: AE example missing a step
The example for the AE table is missing some code. It says we will use the **ae_data** analysis results data and, for brevity, we will subset to AEs with >10% prevalence in the High Dose group. I do not see any code to do this subset. The first chunk of code is:

```
tfrmt(
  # specify columns in the data
  group = AEBODSYS,
  label = AETERM,
  column = c(col2, col1), 
  param = param,
  values = value,
  sorting_cols = c(ord1, ord2)) %>%
  print_to_gt(ae_data2) %>%
  tab_options(
    container.width = 1000
  )

```

… which makes use of **ae_data2**. I assume that ae_data2 is the dataframe which contains the subset mentioned previously (>10% prevalence) … but the user is not shown how to create it.

shannonhaughton:
Fixes #164

### DIFF
--- a/vignettes/examples.Rmd
+++ b/vignettes/examples.Rmd
@@ -229,7 +229,7 @@ For the adverse events (AE) table, we will use the `ae_data` analysis results da
   ungroup %>% 
   filter(pct_high >10) %>% 
   select(-pct_high)
-```
+  ```
 </details>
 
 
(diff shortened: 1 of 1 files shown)


statasaurus (approved):
Wonderful! Thank you

PR #207, opened 2022-07-12
Create examples

shannonhaughton:
Adding examples to exported functions


statasaurus (approved):
🐐 Thank you for adding in all these examples!

R/tfrmt_sigdig.R
```diff
@@ -168,9 +168,12 @@ param_set <- function(...){
 #'           values = value,
 #'           sorting_cols = vars(ord1, ord2, ord3),
 #'           col_plan = col_plan(-starts_with("ord"))) %>%
-#'     print_to_gt(data_labs)
+#'     print_to_gt(labs_data)
```
statasaurus, line 171: I just realized some of these need to be changed to reflect the new data names

statasaurus (changes requested):
You just need to go through and change the datasets from *_data to data_*

shannonhaughton:
@statasaurus I think there was only one case of this, have done a find in files and cant see any more

statasaurus (approved):
Amazing 🐄

Issue #170, opened 2022-06-28
tfrmt_sigdig scoping issue


bzkrouse:
needs to restrict to looking inside the input DF for "grp" - instead it tries to use the numeric grp vector in first line

```
grp <- c(1, 2)
mydf <- tibble(grp = c("one", "two"),
               sigdig = c(1,2))
tfrmt_sigdig(mydf,
             group = grp, 
             label = quo(),
             param_defaults = param_set()) 
```

statasaurus:
Can we add this to the documentation of tfrmt to tell people

Issue #156, opened 2022-06-23
tfrmt does not build correctly if input values have same name as other arguments


bzkrouse:
This is ok:

```
tfrmt(label = group, group = label)
```
 as is this:
```
tfrmt(label = label)
```

but this is not:
```
tfrmt(label = group)
```

In the latter, there is no message to user and tfrmt object is created, but label is empty.

statasaurus:
This will work if group is in quotes 
```{r}
tfrmt(label = "group")
```

PR #221, opened 2022-07-16
Add explanation of argument evaluation. (#170)

thebioengineer:
Hopefully this is clear enough to explain how the arguments evaluate?

We could also address #156 here too if we want?

### DIFF
--- a/R/tfrmt.R
+++ b/R/tfrmt.R
@@ -36,7 +36,7 @@
 #'   passed to it is intended to be an unquoted expression representing a column
 #'   name or an object from the environment. As such, it preferentially uses the
 #'   value from the environment over preserving the entry as an expression. For
-#'   example, if in your environment you have an object "my_object" in your
+#'   example, if you have an object "my_object" in your
 #'   environment with the value "Hello world", and try to create a tfrmt as
 #'   `tfrmt(column = my_object)`, it will take the value of "my_object" over
 #'   assuming the column argument is an unquoted expression and view the entry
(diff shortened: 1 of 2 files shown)


R/tfrmt.R
```diff
@@ -23,7 +23,24 @@
...
+#'   tfrmt allows users to pass `vars`, `quo`, and unquoted expressions to a
+#'   variety of arguments, such as `group`, `label`, `param`, `values`,
+#'   `column`, and `sorting_cols`. Users accustomed to tidyverse semantics
+#'   should be familiar with this behavior. However, there is an important
+#'   behavior difference between tfrmt and normal tidyverse functions. Because
+#'   the data are not a part of tfrmt, it does not know when a value being
+#'   passed to it is intended to be an unquoted expression representing a column
+#'   name or an object from the environment. As such, it preferentially uses the
+#'   value from the environment over preserving the entry as an expression. For
+#'   example, if in your environment you have an object "my_object" in your
```
bzkrouse, line 39:
```suggestion
#'   example, if you have an object "my_object" in your
```

R/tfrmt.R
```diff
@@ -23,7 +23,24 @@
...
+#'   environment with the value "Hello world", and try to create a tfrmt as
+#'   `tfrmt(column = my_object)`, it will take the value of "my_object" over
+#'   assuming the column argument is an unquoted expression and view the entry
+#'   to `column` as "Hello World".
```
bzkrouse, line 43: should we add a line about how to make sure this doesn't happen (i.e. quoting the input)

thebioengineer: That makes sense to add IMO. 
```suggestion
#'   to `column` as "Hello World". To pass "my_object" to tfrmt as a column name, use
#'   quotes around the value: `tfrmt(columnn = "my_object")`.
```

statasaurus (approved):
Perfect!

PR #222, opened 2022-07-18
Layering Vignette - update group

thebioengineer:
Adding to the layering vignette with a light example for `update_group()`


statasaurus (approved):
🐐 This is great! thank you for adding it in

PR #227, opened 2022-07-18
Issue 156

### LINKED ISSUE #156: tfrmt does not build correctly if input values have same name as other arguments

bzkrouse:
resolves #156

### DIFF
--- a/R/tfrmt.R
+++ b/R/tfrmt.R
@@ -27,7 +27,7 @@
 #'
 #'  ## NSE and Argument Evaluation
 #'
-#'   tfrmt allows users to pass `vars`, `quo`, and unquoted expressions to a
+#'   - tfrmt allows users to pass `vars`, `quo`, and unquoted expressions to a
 #'   variety of arguments, such as `group`, `label`, `param`, `values`,
 #'   `column`, and `sorting_cols`. Users accustomed to tidyverse semantics
 #'   should be familiar with this behavior. However, there is an important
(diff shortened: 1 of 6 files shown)


R/tfrmt.R
```diff
@@ -43,6 +43,11 @@
 #'   to `column` as "Hello World". To pass "my_object" to tfrmt as a column name, use
 #'   quotes around the value: `tfrmt(columnn = "my_object")`.
 #'
+#'   - In a similar vein, when supplying variable names that match `tfrmt`
+#'   argument names, it is recommended to put quotes around the value as such:
+#'   `tfrmt(label = "group")`. In this case, the quoting will prevent `tfrmt`
+#'   from assigning its `group` input value to the `label` value.
```
thebioengineer, line 49:
```suggestion
#'   - Additionally, unquoted expressions that match `tfrmt`'s other 
#'   argument names can cause unexpected results. It is recommended 
#'   to put quotes around the value as such:
#'   `tfrmt(label = "group")`. In this case, the quoting will prevent `tfrmt`
#'   from assigning its `group` input value to the `label` value.
```

```suggestion
#'   - Additionally, unquoted expressions that match `tfrmt`'s other argument names can cause unexpected results. 
```
The saying "in a similar vein" is a little colloquial/would potentially require readers to look up that saying is. I think just getting to the point may be more inclusive.

thebioengineer (approved):

statasaurus (approved):

PR #229, opened 2022-07-19
Print to ggplot

shannonhaughton:


R/print_to_ggplot.R
```diff
@@ -0,0 +1,158 @@
...
+  if(!is_tfrmt(tfrmt)){
+    stop("Requires a tfrmt object")
+  }
+
+  if(!is.data.frame(.data)){
+    stop("Requires data")
+  }
+
+  if(is.null(tfrmt$body_plan)==FALSE | is.null(tfrmt$col_plan)==FALSE | is.null(tfrmt$row_grp_plan)==FALSE | is.null(tfrmt$col_style_plan)==FALSE){
+    stop("print_to_ggplot is not currently compatible with Body Plan, Row Group Plan, Column Plan or Column Style Plan. Please remove before continuing.")
```
statasaurus, line 47: A body plan would mean you can't do any rounding. Is there a reason that isn't allowed? It should be applied in `apply_tfrmt` so I don't think it will cause you problems. similarly row_group_plan shouldn't be an issue as long as the element_row_grp_loc is a indented. I think column plan should also be fine.

But, I like the idea of the check. I think it just needs to be more specific like: 
- element_row_grp_loc must be indented is row_group_plan is present 
- col_plan can't include spanning columns 
- column can only have a single column

statasaurus (changes requested):
This looks really good! I think you just need to make a few changes on what is and isn't allowed. I am happy to talk about this and help you test stuff if you need

R/print_to_ggplot.R
```diff
@@ -0,0 +1,180 @@
...
+  .data<-apply_grp_ggplot(.data,tfrmt) %>%
+    mutate(y=n():1)
+
+  # handle cases for "..tfrmt_row_grp_lbl pivoting
+  if("..tfrmt_row_grp_lbl" %in% names(.data)){
+  # reshape data for ggplot
+  data <- .data %>%
+    pivot_longer(-c(as_label(tfrmt$label),"y","..tfrmt_row_grp_lbl"),names_to = "column",values_to="value") %>%
+    mutate(value=if_else(`..tfrmt_row_grp_lbl`==TRUE,"",value))
+  }else(
```
statasaurus, line 116: The else here needs curly brakets

statasaurus (changes requested):
So this doesn't consistently work. 
```tbl_dat <- tibble(
  # grp = "1 to 7 days after treatment",
  ae = rep(c("Fever", "Malaise", "Local Allergic Reaction"), each = 4),
  trt = rep(rep(c("TRT A", "Placebo"), each = 2), 3),
  param = rep(c("n", "pct"), 6),
  value = c(60, 7.7, 26, 3.3, 183, 23.5, 89, 11.4, 94, 12, 55, 7)
)


tfrmt_n_pct() %>% 
  tfrmt(label = ae,
        column = trt,
        param = param, 
        values = "value") %>% 
  print_to_ggplot(tbl_dat)
```
breaks it

statasaurus:
The reason this code was breaking stuff is cause you use data$label and data$column a lot. But if you label column isn't called label everything falls apart

thebioengineer (approved):
Looks good, added my suggestions

statasaurus (approved):
thanks everyone this looks great

PR #235, opened 2022-07-19
correct group/labels with incomplete body_plan

### LINKED ISSUE #177: When the formatting is slightly off the column order gets messed up

bzkrouse:
resolves #177  (I think)

When the body_plan is incomplete, specifically when a frmt_combine is left out, this means we get list-columns after the pivot_wider. Adding some tweaks to make the row group relabeling work in this scenario.


tests/testthat/test-utils_tfrmt.R
```diff
@@ -264,3 +264,47 @@ test_that("Test body_plan missing", {
                  pivot_wider(names_from = column, values_from = val))
 })
 
+
+test_that("incomplete body_plan",{
```
thebioengineer, line 268:
```suggestion
test_that("incomplete body_plan where params share label",{
```

R/apply_row_grp_plan.R
```diff
@@ -244,25 +245,33 @@ combine_group_cols <- function(.data, group, label, element_row_grp_loc = NULL){
             select(!!!top_grouping, !!label) %>%
             mutate(!!label := !!last(group)) %>%
             distinct()
+
+          new_row_other_cols <- lone_dat %>%
```
thebioengineer, line 249: We should add a comment indicating what this is doing. This is getting all the _non_ group/label cols. One question is you are also not grabbing the numeric columns. why is that? numeric cols won't be converted to blanks in the gt. So maybe being consistent and just returning NA for all columns and letting downstream handle that may be better?

bzkrouse: I just wanted to avoid having to conditionally convert character to "" and numeric leaving as NA. But we have to conditionally handle list columns and character columns anyway, so I'm fine being even more explicit

bzkrouse: I've added a comment and moved the discussion about returning NAs to a separate issue/PR

statasaurus (approved):
Looks good thank you for adding the comment it helps

PR #241, opened 2022-07-20
Tfrmt pivot wider message

thebioengineer:
Make a message when there is the potential to pivot wider and generate a listcol. It tries to guess what the frmt_structure that is missing may be, but this is a "simple" implementation. We could try to get it to collapse group_val/label_val, but I wanted to run the concept first:

```r
Multiple param listed for the same group/label. Following frmt_structures may be missing from the body_plan or the order may need to be changed:
- `frmt_structure(group_val = "grp1_1", label_val = "my_label_1", frmt_combine("{N}, {mean}",N = frmt("xx"), mean = frmt("xx")))`
- `frmt_structure(group_val = "grp1_1", label_val = "my_label_2", frmt_combine("{N}, {mean}",N = frmt("xx"), mean = frmt("xx")))`
- `frmt_structure(group_val = "grp1_2", label_val = "my_label_1", frmt_combine("{N}, {mean}",N = frmt("xx"), mean = frmt("xx")))`
- `frmt_structure(group_val = "grp1_2", label_val = "my_label_2", frmt_combine("{N}, {mean}",N = frmt("xx"), mean = frmt("xx")))`
```


R/apply_tfrmt.R
```diff
@@ -194,33 +194,61 @@ remove_empty_layers <- function(x, nlayers = 1){
...
+          lbl = !!tfrmt$label,
+          param_vals = .data$param_list
+          )
+      ) %>%
+      pull(.data$suggested_frmt_struct) %>%
+      paste0("- `",.,"`",collapse = "\n")
+
+
+   inform(
+     "Multiple param listed for the same group/label. Following frmt_structures may be missing from the body_plan or the order may need to be changed:",
```
statasaurus, line 232: This is a bit too wide to read can you put more line breaks in

statasaurus (changes requested):
looks pretty good. I think we just need a tidy up the messages a bit more

R/apply_tfrmt.R
thebioengineer: done

bzkrouse:
> looks pretty good. I think we just need a tidy up the messages a bit more

I like the idea of the suggestion but at the same time it is hard to anticipate what they want and this gets very verbose with a table with many group/label values. What if the msg was more general and geared to the params that end up in cells together?

thebioengineer:
> > looks pretty good. I think we just need a tidy up the messages a bit more
> 
> I like the idea of the suggestion but at the same time it is hard to anticipate what they want and this gets very verbose with a table with many group/label values. What if the msg was more general and geared to the params that end up in cells together?
> 
> i.e. "consider adding a frmt_structure for the following sets of params: [list params]" and we could include 1 example of how to do that but not necessarily cover all values

I think we need to provide more than just params, so that users can be aware of what went wrong & guidance on fixing it. I agree that it could get unwieldy, but since this behavior is not wanted and we are just permitting it (not defining what should be frmt_combine) making the message very loud I don't think is a bad thing.

statasaurus (approved):
This looks good. In the future we might want to add in a test with back ticks in it. I don't think there should be a problem, but if anything would cause a problem they would

PR #240, opened 2022-07-20
Rowlbl added rows

bzkrouse:
see #239

PR #249, opened 2022-07-22
Frmt combine check ehh

thebioengineer:
Hey, I couldnt remember if I was supposed to add the replace_na or not (#242). It led me down to find a bug where label was being removed if it didnt have the first param in a frmt_combine - obviously not good. 

~~This branch will _fail_ for now, because it needs the fixes in #240 applied.~~


R/apply_frmt_methods.R
```diff
@@ -212,7 +223,10 @@ apply_frmt.frmt_combine <- function(frmt_def, .data, values, mock = FALSE, param
     ) %>%
     mutate(
       .is_all_missing =  all_missing(fmt_param_vals,.)
-    )
+    ) %>%
+    ## after .is_all_missing so that can be tabulated first
+    replace_na(missing_param_replacements)
```
statasaurus, line 228: This needs an if around it. Cause at the moment it will cover up NA when missing isn't set

statasaurus (changes requested):
The replace_na is a bit on the agressive side right now. I think you need an if around it. Otherwise it replaces NA's when missing isn't specified

R/apply_frmt_methods.R
statasaurus: There is also an issue when a frmt_when is used. Do we need to add a missing option to frmt_when as well?

thebioengineer: That is simple enough to sort out. for some reason I thought we converted NA's into "". WRT frmt_when including a missing case, that seems reasonable. Alternatively for the "default" case, it could be expected to drop the missing value in there.

tests/testthat/test-utils_tfrmt.R
```diff
@@ -186,7 +189,8 @@ test_that("Check apply_tfrmt for mock data",{
     )
   )
   mock_dat <- make_mock_data(plan, .default = 1, n_col = 1) %>%
-    apply_tfrmt(plan, mock =TRUE)
+    quietly(apply_tfrmt)(plan, mock =TRUE) %>%
```
bzkrouse, line 192: Since you used quietly() was this supposed to return a message?

thebioengineer: it is returning the message about the ways to correct multiple params per group/label/col. That isnt the point of the test, so I wanted it to be quiet. I find that unsuppressed non-relevant messages clutter up the ability to read and understand the testthat output

tests/testthat/test-utils_tfrmt.R
```diff
@@ -217,8 +221,15 @@ test_that("Check apply_tfrmt for mock data",{
...
 
-  expect_message(mock_dat %>% apply_tfrmt(plan, mock =TRUE),
-                "Mock data contains more than 1 param per unique label value. Param values will appear in separate rows.")
+  make_mock_dat_message <- mock_dat %>%
+    apply_tfrmt(plan, mock =TRUE) %>%
+    capture_messages()
+
+  ## capturing second message
+  expect_equal(
+    make_mock_dat_message[2],
```
bzkrouse, line 230: I see that we are also keeping the regular frmt_combine msg for mocks. 
If someone creates mocks w/o data, they are unable to say if multiple params live on the same row, in different columns. This would be unrelated to frmt_combine but you'd still receive the frmt_combine suggestion. So, in the case of mocks should we have more general messaging?

thebioengineer: That is a good point!

R/apply_frmt_methods.R
```diff
@@ -242,31 +252,37 @@ apply_frmt.frmt_combine <- function(frmt_def, .data, values, mock = FALSE, param
...
-      left_join(
-        .tmp_data_fmted,
-        by = map_chr(merge_group, as_label)
-      )
+    .data <- .data %>%
+      select(-!!values)
   }
-  out
+
+  merge_group <- map(
```
bzkrouse, line 268: is this duplicate code for line 254-260?

tests/testthat/test-utils_tfrmt.R
bzkrouse: yep that makes sense! i was just running it locally and there was no message captured by the quietly.

bzkrouse (changes requested):
this looks and is working fab!! 
 I put a couple of suggestions for edits in the comments. Let me know your thoughts.

R/apply_frmt_methods.R
thebioengineer: derp. Yes. corrected

tests/testthat/test-utils_tfrmt.R
thebioengineer: because we allow users to provide their own data, this is a little difficult to blindly not return the message whether they need to correct the bodyplan for multiple param/cell

thebioengineer: I just set it to not return those messages when its running a mock, but we may want to think more about it going forward

bzkrouse (approved):
Thanks for the changes! ⭐ 🥇

R/apply_frmt_methods.R
```diff
@@ -214,6 +214,17 @@ apply_frmt.frmt_combine <- function(frmt_def, .data, values, mock = FALSE, param
       .is_all_missing =  all_missing(fmt_param_vals,.)
     )
 
+  missing_param_replacements <-
+    map(fmt_param_vals, ~ frmt_def$fmt_ls[[.x]]$missing) %>%
```
statasaurus, line 218: If you just use 
```
map(frmt_def$fmt_ls,  ~.$missing) 
```
The come out named

statasaurus (approved):
Looks great! Thanks for the updates

PR #269, opened 2022-08-05
rename tfrmt_sigdig input data argument 

### LINKED ISSUE #258: possibly rename tfrmt_sigdig `data` argument
...to be more specific to tfrmt_sigdig, i.e. metadata, data_sigdig, sigdig_df

bzkrouse:
resolves #258

### DIFF
--- a/R/tfrmt_sigdig.R
+++ b/R/tfrmt_sigdig.R
@@ -104,19 +104,20 @@ param_set <- function(...){
 #' Create tfrmt object from significant digits spec
 #'
 #' This function creates a tfrmt based on significant digits specifications for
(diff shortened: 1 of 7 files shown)


thebioengineer:
tfrmt_sigdig is also found in the layering vignette. 

Additionally, I thought we were going to do something like `sigdig_df`? Just because its not actually `data`, i am wary of using that as the first part of the argument name. Additionally, due to partial arg matching, the layer vignette was still able to run.

bzkrouse:
@thebioengineer  Done and done. I had checked the templates vignette but forgot to check layer. Thanks!!

statasaurus (approved):
Sorry this took so long to pull in but it look great!

PR #282, opened 2022-08-12
move handling of "noprint" to apply_row_grp_lbl

### LINKED ISSUE #281: move handling of label_loc = "noprint" to apply_row_grp_lbl
it's currently in apply_col_plan

bzkrouse:
resolves #281


R/apply_row_grp_plan.R
```diff
@@ -100,22 +100,32 @@ apply_row_grp_lbl <- function(.data, element_row_grp_loc, group, label = NULL, .
...
-        group_by(!!group[[1]])
-    } else { # Using the grouping in gt, but needs to drop all groups in label
-      add_ln_df <- add_ln_df %>%
-        group_by(!!group[[1]]) %>%
-        select(-c(!!!group[-1]))
+    # Either drop group columns ("no print"), or format them w/ label
+
+    if (element_row_grp_loc$location=="noprint"){
+
+      add_ln_df <- .data %>% select(-c(!!!group))
```
thebioengineer, line 107: Do we need to worry about cases where group isn't defined in the tfrmt?

bzkrouse: Great question. We've been leaning on `tentative_process` to catch that for several of these functions. So if there is no tfrmt$group, this function won't be run at all.

thebioengineer: fair fair. I'm fine then with going forward. It may be a discussion point on how much we should rely on tentative process to protect the code vs building in safeguards.

thebioengineer (approved):

PR #283, opened 2022-08-12
Processed tfrmt obj

thebioengineer:
somewhat dependent on #282 (and will need to make updates to reflect that)

updates to processing of data with tfrmt to allow appending of an attribute to include the original and renamed data columns. this is for use by downstream code that generates the gt and adds "big N"


statasaurus (changes requested):
While this is good. I am going to refactor it to include the big n stuff and then I will pull into main

PR #287, opened 2022-08-18
add a line break to help with formatting, change val for align to p_l…

shannonhaughton:
…ow and p_high to stop error


statasaurus (approved):
Looks great!

PR #289, opened 2022-08-18
Big n2

statasaurus:


R/apply_col_plan.R
```diff
@@ -288,6 +304,22 @@ split_data_names_to_df <- function(data_names, preselected_cols, column_names){
 
 }
 
+#' Combines the split out data.frame of the potential columns and their spans into a named vector.
+#'
+#'
+#' @param split_data_names tibble containing a two columns for every
```
bzkrouse, line 310:
```suggestion
#' @param split_data_names tibble containing two columns for every
```

vignettes/big_ns.Rmd
```diff
@@ -0,0 +1,177 @@
...
+    NA,         NA,  "column cols",  "cols 1,2",   NA  ,      "bigN",    12,
+    NA,         NA,  "column cols",  "col 4"   ,   "col4"  ,  "bigN",    6,
+    NA,         NA,             NA,          NA,   "mycol3",  "bigN",    6
+) %>% 
+  bind_rows(with_big_n_data)
+
+print_to_gt(n_span_tfrmt, .data = with_more_big_n_data)
+```
+ 
+As a final note, column names for the big N's should match the ARD column names not any future names. So regardless, if column names get changed in the `col_plan`, the column name for the big N's should reflect the column name of that column you want it to go in the ARD. The big N's will still be added if names do get updated.
```
bzkrouse, line 162:
```suggestion
As a final note, column names for the big N's should match the ARD column names, rather than any future names. Regardless of any renaming that happens in the `col_plan`, the column name for the big N's should be consistent with column values in the ARD. The big N's will still be added if names do get updated in the `col_plan`.  
```

R/big_n.R
```diff
@@ -0,0 +1,111 @@
...
+      param_val = param_val,
+      n_frmt = n_frmt
+    ),
+    class = c("big_n_structure","structure")
+  )
+}
+
+
+
+apply_big_n_df <- function(col_plan_vars, columns, value, big_n_df){
```
bzkrouse, line 35: consider adding roxygen for these internal functions

bzkrouse (approved):
Couple of suggested edits, which you can take or leave. Otherwise good to go!

PR #292, opened 2022-08-22
Values to value

shannonhaughton:


shannonhaughton:
The word value doesn't look real anymore

statasaurus (approved):
Looks great thanks for changing this everywhere

PR #297, opened 2022-08-25
Element col char widths

thebioengineer:
Change behavior to be setting col widths based on number of characters rather than actual px/inches/etc.

Marking as draft so we can track progress/make obvious the differences.


thebioengineer:
I would like to discuss when to do the splitting vs alignment with yall. I have the code to do every piece, just need to sort out order of operations with yall!

PR #314, opened 2022-09-15
Format warning

### LINKED ISSUE #299: Improve warning for unformatted values
It would be good to combine the options rather than give an option for every label type which is a bit overwhelming

shannonhaughton:
fixed #299


statasaurus (approved):
Thanks for updating this. I think it will be a bit easier to read

PR #315, opened 2022-09-15
adding error message for if element_col is missing col argument

### LINKED ISSUE #313: Improve error when element_col is missing the col argument
image
The issues was just col was missing an argument in element_col

shannonhaughton:
fixed #313


statasaurus (approved):
Thanks for adding in this error message I think it will help

PR #321, opened 2022-10-07
Order var align

### LINKED ISSUE #320: Add message where order variables are causing data to be printed on two separate lines
image
image

shannonhaughton:
Fixed #320


R/apply_tfrmt.R
```diff
@@ -34,6 +34,27 @@ apply_tfrmt <- function(.data, tfrmt, mock = FALSE){
       fail_desc = "Failure while aligning data values"
     )
 
+if(is_empty(tfrmt$sorting_cols)==FALSE){
```
statasaurus, line 37: Rather than adding this to the already rather long function that is apply_tfrmt can you break it out into its own function

statasaurus:
Also, if you break the code out to its own function it will be easier to test. You can just make a good dataset and a bad one and the former shouldn't error and the latter should

R/apply_tfrmt.R
shannonhaughton: done

R/apply_tfrmt.R
```diff
@@ -5,7 +5,7 @@
 #' @param mock Logical value is this is for a mock or not. By default `FALSE`
 #'
 #' @return formatted tibble
-#'
+#' @importFrom dplyr n_distinct
```
bzkrouse, line 8: can you move this down to above the helper function that uses it?

R/apply_tfrmt.R
```diff
@@ -361,3 +364,39 @@ frmt_struct_string <- function(grp, lbl, param_vals){
...
+    }else{
+      order_check <- .data %>%
+        group_by(!!tfrmt$label) %>%
+        mutate(n1=n_distinct(!!tfrmt$label,!!!tfrmt$sorting_cols),
+               n2=n_distinct(!!tfrmt$label))
+    }
+
+    # print warning if the number of lines printed over is greater than 1
+    if(sum(order_check$n1)>nrow(order_check) & all(order_check$n1 == order_check$n2)==FALSE){
+      message("Note: Some row labels have values printed over more than 1 line. ")
```
bzkrouse, line 398: Is there a need to keep this as 2 messages? if not, you can collapse it down and use "\n" to separate the lines.

R/apply_tfrmt.R
```diff
@@ -361,3 +364,39 @@ frmt_struct_string <- function(grp, lbl, param_vals){
...
+      message("This could be due to incorrect order variables. Each row in your output table should have only one order var combination assigned to it.")
```
bzkrouse, line 399: I imagine it will be clearer for users if we stick to "sorting cols" rather than order vars per the ?tfrmt argument/documentation.

bzkrouse (changes requested):
Looks great! Just a couple of small comments/suggestions!

statasaurus (approved):

PR #336, opened 2022-10-24
Updating documentation stuff

statasaurus:


thebioengineer (approved):
Only problem I have is deleting tlang.Rproj instead of renaming to tfrmt.Rproj. other than that, its gtg.

Approving to prevent being a blocker, but please correct that.

PR #339, opened 2022-10-25
Add missing "returns" & new test

thebioengineer:
Noticed that a number of our exported functions don't have documented returns, so updated that. 

also added a test for tentative_process since it wasn't explicitly tested afaik for failing functions, etc.


statasaurus (approved):
Thanks for adding in these small updates

PR #344, opened 2022-11-14
:chicken: Change element_col to col_style_structure

statasaurus:


thebioengineer (approved):
I think it looks good, updates to vignette are clear to me.

PR #346, opened 2022-12-15
Fix for dev dplyr

DavisVaughan:
This PR makes your package compatible with the next version of dplyr:

- `if_else()` now propagates names so we need to unname where they aren't expected
- `na_if()` now coerces `y` to the type of `x` so we now need to convert `x` to character early in a place
- You had a place where `str_detect()` could return >1 values inside an `if()` so I wrapped that in `any()`
- `case_when()` and `if_else()` now retain the type of matrices rather than coercing them to vectors, so they need to be coerced to vector early
- `distinct()` now retains the ordering supplied by the user, so I've reordered with a `select()` early on to ensure this is compatible with both dev and CRAN dplyr
- There are a few places where I switched to `expect_no_message()` or `expect_snapshot()` due to new warnings that appear in the join functions. The join functions in dplyr (like `left_join()`) now return a warning by default when a row in `x` matches _multiple_ rows in `y`. While this is typical SQL behavior, it is often unexpected during data analysis (many people don't even know it is possible), so we've decided to make this a warning. In dplyr 1.1.0, you silence this warning with `multiple = "all"`.

We plan to submit dplyr 1.1.0 on January 27th.

This should be compatible with both dev and CRAN dplyr. It would help us out if you could go ahead and send a patch version of your package to CRAN ahead of time! Thanks!

### DIFF
--- a/R/apply_footnote_meta.R
+++ b/R/apply_footnote_meta.R
@@ -119,7 +119,7 @@ get_col_loc <- function(footnote_structure, .data, col_plan_vars, columns){
       if(last(col_str) == span_lvl){
         col_loc <- unite_df_to_data_names(col_loc_df, preselected_cols = c(), column_names = col_str)
         if(!is.null(names(col_loc))){
-          col_loc <- if_else(names(col_loc) != "", names(col_loc), col_loc)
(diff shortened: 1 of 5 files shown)


statasaurus (approved):
Thank you so much

PR #380, opened 2023-07-07
add multi alignment

bzkrouse:


statasaurus (changes requested):
I would add this update to a new section of the news file and up version the package

statasaurus (approved):

PR #390, opened 2023-08-10
Page plan json

bzkrouse:


statasaurus (approved):
so this looks fine, but there is a slight issue with the JSON files generally. Body plan despite having a list of frmt_structures doesn't name that list in the json. Then footnote_plan calls it 'struct_list' like here but row_grp_plan calls it 'struct_ls'. So I just think we want to be consistent to make it easier for people making the json files

bzkrouse:
@statasaurus Good catch. The naming comes from how the functions are defined originally. For plans that ONLY take a series of structures (like body_plan) we have no need for the struct_ls/struct_list. But for other plans, I corrected them to all use `struct_list`

PR #425, opened 2024-01-18
Use similar logic in page_plan and row_grp_plan

### LINKED ISSUE #381: row_grp_structure with multiple group variables should assume NULL if not specified
suppose there are 2 group variables, grp1 and grp2, and the row_grp_plan only references one of them:

`row_grp_structure(group_val = list(grp1 = ".default"), ...)`

Current behavior is essentially to assume grp2 = ".default", meaning there will be 1 block per unique value of grp1 and grp2 combined. Desired behavior would be to assume grp2 = NULL, and therefore the blocks will only be based on unique values of grp2.

Linking to #374 as this is the same logic we want to use for page_structure to achieve the appropriate "group by" logic. Note that frmt_structure and footnote_structure should assume variables that are unaccounted for in the list are set to ".default" as these are purely for filtering on values rather than performing by groups.

bzkrouse:
closes #381


karima662 (approved):
Looks good 👍

PR #428, opened 2024-02-01
create check_plan()

karima662:
- check_plan function to be used in following format: `check_plan(tfrmt_object, "plan")`
- currently used within tfrmt() function
- stops and presents error message when non-null plan not supplied with plan function

The following can be used to test:

```r
dat <- tribble(
  ~group,     ~label,  ~my_col,    ~parm, ~val,
  "g1", "rowlabel1",  "col1"  ,  "value",    1,
  "g1", "rowlabel1",  "col2"  ,  "value",    1,
  "g1", "rowlabel1",  "mycol3",  "value",    1,
  "g1", "rowlabel1",  "col4"  ,  "value",    1,
  "g1", "rowlabel1",  "mycol5",  "value",    1,
  "g1", "rowlabel2",  "col1"  ,  "value",    2,
  "g1", "rowlabel2",  "col2"  ,  "value",    2,
  "g1", "rowlabel2",  "mycol3",  "value",    2,
  "g1", "rowlabel2",  "col4"  ,  "value",    2,
  "g1", "rowlabel2",  "mycol5",  "value",    2,
  "g2", "rowlabel3",  "col1"  ,  "value",    3,
  "g2", "rowlabel3",  "col2"  ,  "value",    3,
  "g2", "rowlabel3",  "mycol3",  "value",    3,
  "g2", "rowlabel3",  "col4"  ,  "value",    3,
  "g2", "rowlabel3",  "mycol5",  "value",    3)

tfrmt(
  group = group,
  label = label,
  param = parm,
  value = val,
  column = my_col,
  # body_plan function supplied
  body_plan = body_plan(frmt_structure(group_val = ".default", label_val = ".default", frmt("x")))
  ) %>% 
  print_to_gt(dat)

tfrmt(
  group = group,
  label = label,
  param = parm,
  value = val,
  column = my_col,
  # body_plan function *not* supplied
  body_plan = frmt_structure(group_val = ".default", label_val = ".default", frmt("x"))
  ) %>% 
  print_to_gt(dat)
```

### DIFF
--- a/R/tfrmt.R
+++ b/R/tfrmt.R
@@ -197,6 +197,15 @@ tfrmt <- function(
     class = c("tfrmt")
   )
 
+  # check non-null plan parameters are supplied with plan function
+  check_plan(new_tfrmt, "body_plan")
+  check_plan(new_tfrmt, "col_plan")
+  check_plan(new_tfrmt, "col_style_plan")
+  check_plan(new_tfrmt, "row_grp_plan")
+  check_plan(new_tfrmt, "footnote_plan")
+  check_plan(new_tfrmt, "page_plan")
+
+
   if(!missing(tfrmt_obj)){
     new_tfrmt <- layer_tfrmt(
       tfrmt_obj,
@@ -213,6 +222,7 @@ tfrmt <- function(
   check_footnote_plan(new_tfrmt)
 
   new_tfrmt
+  #browser()
 
 }
 
--- a/R/tfrmt_checks.R
+++ b/R/tfrmt_checks.R
@@ -284,3 +284,24 @@ check_footnote_plan <- function(x){
 
 }
 
+
+#' Check if the plan parameter is consistent in tfrmt
+#' @noRd
+#' @param tfrmt_object tfrmt object to be checked
+#' @param plan plan parameter, e.g., col_style_plan
+#' @importFrom rlang abort inherits
+#'
+#'
+check_plan <- function(tfrmt_object, plan) {
+  # extract the plan element from the tfrmt_object
+  plan_element <- tfrmt_object[[plan]]
+
+  # check if the user supplied a value to the plan parameter
+  if (!is.null(plan_element)) {
+    # Check if the plan element inherits the plan attribute
+    if (!inherits(plan_element, plan)) {
+      stop("Check if ", plan, " parameter has been supplied ", plan, "() function")
+    }
+  }
+}
+


R/tfrmt.R
```diff
@@ -213,6 +222,7 @@ tfrmt <- function(
   check_footnote_plan(new_tfrmt)
 
   new_tfrmt
+  #browser()
```
bzkrouse, line 225:
```suggestion 
```

R/tfrmt_checks.R
```diff
@@ -284,3 +284,24 @@ check_footnote_plan <- function(x){
 
 }
 
+
+#' Check if the plan parameter is consistent in tfrmt
+#' @noRd
+#' @param tfrmt_object tfrmt object to be checked
+#' @param plan plan parameter, e.g., col_style_plan
+#' @importFrom rlang abort inherits
```
bzkrouse, line 292:
```suggestion
#' @importFrom rlang abort
```

bzkrouse: inherits is a base function

R/tfrmt_checks.R
```diff
@@ -284,3 +284,24 @@ check_footnote_plan <- function(x){
 
 }
 
+
+#' Check if the plan parameter is consistent in tfrmt
+#' @noRd
+#' @param tfrmt_object tfrmt object to be checked
+#' @param plan plan parameter, e.g., col_style_plan
```
bzkrouse, line 291:
```suggestion
#' @param plan name of plan, e.g., "col_style_plan"
```

R/tfrmt_checks.R
```diff
@@ -284,3 +284,24 @@ check_footnote_plan <- function(x){
...
+#'
+check_plan <- function(tfrmt_object, plan) {
+  # extract the plan element from the tfrmt_object
+  plan_element <- tfrmt_object[[plan]]
+
+  # check if the user supplied a value to the plan parameter
+  if (!is.null(plan_element)) {
+    # Check if the plan element inherits the plan attribute
+    if (!inherits(plan_element, plan)) {
+      stop("Check if ", plan, " parameter has been supplied ", plan, "() function")
```
bzkrouse, line 303: Can you please revise to be a bit clearer about what is wrong, e.g. that an invalid *_plan is supplied to the argument

bzkrouse (changes requested):
thanks for this, looking great. Aside from the suggestions proposed to the code, can you also please:

(1) add your unit test
(2) also check big_n (you may need to write another function for this, as big_n expects a big_n_structure)

R/tfrmt_checks.R
karima662: Does the new error message in (https://github.com/GSK-Biostatistics/tfrmt/pull/428/commits/b9abc09dd3ca6a937e6cc504588eab8c70fbcd5e) sound better?

image

bzkrouse: I think just "Invalid input supplied to the `col_style_plan` parameter. Please supply a `col_style_plan()`.

R/tfrmt_checks.R
```diff
@@ -284,3 +284,24 @@ check_footnote_plan <- function(x){
...
+#'
+check_plan <- function(tfrmt_object, plan) {
+  # extract the plan element from the tfrmt_object
+  plan_element <- tfrmt_object[[plan]]
+
+  # check if the user supplied a value to the plan parameter
+  if (!is.null(plan_element)) {
+    # Check if the plan element inherits the plan attribute
+    if (!inherits(plan_element, plan)) {
+      stop("Invalid *_plan function has been supplied to the ", plan, " parameter")
```
bzkrouse: Apologies for not catching this on my previous review- it would be good to use rlang::abort instead, as this will allow the error to show up as "Error in tfrmt()" rather than "Error in check_plan()". Two changes you will have to make: 
(1) pass the parent environment as another argument in the function
(2) change to rlang::abort, passing the environment into its `call` argument 

There is a bit of this happening in some of the other helpers but it might be easier to talk through it together. more info here: https://rlang.r-lib.org/reference/topic-error-call.html#passing-the-user-context

tests/testthat/test-tfrmt_checks.R
```diff
@@ -40,3 +40,77 @@ test_that("Testing error messages for missing argument col_style_structure",{
     ),
     "Column element is missing from col_style_structure[.] Note: col here refers to the values within the column variable in your data, rather than the variable name itself")
 })
+
+
+
+test_that("Testing error message for invalid input to plan parameters, specifically for col_style_plan",{
```
bzkrouse, line 46: for both of these new tests, can you please add the expected error as well? Similar to the test above

bzkrouse (changes requested):
This is looking awesome - just 1 comment for the tests!

tests/testthat/test-tfrmt_checks.R
karima662: Due to the revision made as a result of https://github.com/GSK-Biostatistics/tfrmt/pull/428#discussion_r1480020661 , expected error is:
image
When this message is defined as expected in test (line 73):
image
The test fails:
image

Hence, a workaround was to use tryCatch() instead of expect_error() - as done in https://github.com/GSK-Biostatistics/tfrmt/pull/428/commits/f7c0b5a49c4aed3dd0b5c22d13d4b24c607d3cc2

bzkrouse (changes requested):
By default, expect_error does regular expression matching. Since our error message contains some special characters, we have to set fixed=TRUE to do an exact match rather than regex. Therefore you can simplify to:
```
expect_error(
   tfrmt(...),
   "message",
   fixed=TRUE
)
```

Does that make sense?

bzkrouse (approved):

PR #442, opened 2024-03-04
NA label custom msg

karima662:
edited `apply_row_grp_lbl` to:

- store label column values
- check if any NAs present
- output custom message if true

### DIFF
--- a/R/apply_row_grp_plan.R
+++ b/R/apply_row_grp_plan.R
@@ -96,6 +96,15 @@ apply_row_grp_struct <- function(.data, row_grp_struct_list, group, label = NULL
 #' @importFrom rlang !!!
 apply_row_grp_lbl <- function(.data, element_row_grp_loc, group, label = NULL, ...){
 
+  # store values of label column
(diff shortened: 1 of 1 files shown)


R/apply_row_grp_plan.R
```diff
@@ -96,6 +96,15 @@ apply_row_grp_struct <- function(.data, row_grp_struct_list, group, label = NULL
 #' @importFrom rlang !!!
 apply_row_grp_lbl <- function(.data, element_row_grp_loc, group, label = NULL, ...){
 
+  # store values of label column
+  lbl_col <- rlang::eval_tidy(label, .data)
```
bzkrouse, line 100: please move this to @importFrom rlang eval_tidy, and remove the rlang::

bzkrouse (changes requested):
This is just what we need thank you!! Just a tiny change to do and we should be good!

R/apply_row_grp_plan.R
karima662: see latest commit :)

bzkrouse (approved):
thank you!!

PR #444, opened 2024-03-06
Json example tfrmts

karima662:
addressing #427


karima662:
The json for the ae example is currently not working

bzkrouse:
general comment: can you please rename the json files to be `tfrmt_*.json` to match the convention of our example data, which is `data_*`? Thank you!!

Issue #446, opened 2024-03-13
add test for NA in label column error


karima662:
add test for changes made in #442 

currently have the following code, but the test fails:

```r
test_that("Row group plans when label column contains NA - error check",{

  # create data
  data_ae2 <- data_ae %>%
    group_by(AEBODSYS, AETERM) %>%
    mutate(pct_high = value[col2=="Xanomeline High Dose" & param=="pct"]) %>%
    ungroup %>%
    filter(pct_high >10) %>%
    select(-pct_high)

  data_ae2$AETERM <- ifelse(data_ae2$AETERM == "ANY BODY SYSTEM", NA, data_ae2$AETERM)


  # create tfrmt and print - expect error
  expect_error({
    tfrmt(
      group = AEBODSYS,
      label = AETERM,
      column = c(col2, col1),
      param = param,
      value = value,
      sorting_cols = c(ord1, ord2),
      body_plan = body_plan(
        frmt_structure(group_val = ".default", label_val = ".default",
                       frmt_combine("{n} {pct}",
                                    n = frmt("XXX"),
                                    pct = frmt_when(
                                      "==100" ~ "",
                                      "==0" ~ "",
                                      TRUE ~ frmt("(xx.x %)")))),
        frmt_structure(group_val = ".default", label_val = ".default",
                       AEs = frmt("[XXX]")),
        frmt_structure(group_val = ".default", label_val = ".default",
                       pval = frmt_when(">0.99" ~ ">0.99",
                                        "<0.001" ~ "<0.001",
                                        "<0.05" ~ frmt("x.xxx*"),
                                        TRUE ~ frmt("x.xxx", missing ="--")))
      ),
      col_plan = col_plan(
        -starts_with("ord")
      )) %>%
      print_to_gt(., data_ae2) %>% tab_options(container.width = 1000)

  },
  paste("Unable to to apply apply_row_grp_lbl.",
        "Reason: `label` column AETERM contains NA values. For group-level summary data, `label` and the relevant `group` values should match.",
  sep = "\n"),
  )
})

```

karima662:
solution - change expect_error to expect_message

PR #448, opened 2024-03-13
addressing #446

karima662:


tests/testthat/test-apply_row_grp_plan.R
```diff
@@ -955,3 +955,92 @@ test_that("Row group plan indenting handles factor variables", {
...
+    "grp1_1", "grp2_1", "my_label_2",
+    "grp1_1", "grp2_2", "my_label_1",
+    "grp1_1", "grp2_2", "my_label_2",
+    "grp1_2", "grp2_1", "my_label_1",
+    "grp1_2", "grp2_1", "my_label_2",
+    "grp1_2", "grp2_2", "my_label_1",
+    "grp1_2", "grp2_2", "my_label_2",
+  ) %>%
+    mutate(
+      trtA = rep("xx (xx%)", 8) %>% as.list(),
```
bzkrouse, line 1026: we don't need the list-columns for this test. can you remove the `%>% as.list() ` from the creation of trtA/B/C?

tests/testthat/test-apply_row_grp_plan.R
```diff
@@ -955,3 +955,92 @@ test_that("Row group plan indenting handles factor variables", {
...
+
+  sample_grp_plan <- row_grp_plan(
+    row_grp_structure(group_val = ".default", element_block(post_space =" ")),
+    label_loc = element_row_grp_loc(location = "indented")
+  )
+
+  # expect error message
+  expect_error({
+    auto_test_listcols <- apply_row_grp_lbl(mock_multi_grp, sample_grp_plan$label_loc,group = vars(grp1, grp2), label = sym("my_label")) %>%
+      remove_grp_cols(sample_grp_plan$label_loc,group = vars(grp1, grp2),label = sym("my_label"))
```
bzkrouse, line 1041: the `remove_grp_cols` line is also not needed for this test, can you please remove? :) It should be something like this:

```
  # expect error message
  expect_error({
    apply_row_grp_lbl(mock_multi_grp, sample_grp_plan$label_loc,group = vars(grp1, grp2), label = sym("my_label"))
  },
  paste("`label` column my_label contains NA values. For group-level summary data, `label` and the relevant `group` values should match.")
  )
```

bzkrouse (changes requested):
Looking great, just suggesting a couple changes for the second test to remove some pieces that are not needed. LMK if any questions!!

bzkrouse (approved):
Great!! thank you!!

PR #449, opened 2024-03-14
Json fix

bzkrouse:


bzkrouse:
Co-developed and tested with @karima662.

PR #471, opened 2024-08-28
small fixes for creating mock data

### LINKED ISSUE #467: incorporate values from `col_style_plan` in creation of mock data
as print_mock_gt and print_to_gt utilize much of the same logic, they present the same messages/errors if any of the plans fail due to values not being available in the dataset. For mocks, we should be quieter about any of these failures, or avoid them altogether. 
This may involve changing the make_mock_data functionality too

bzkrouse:
resolves #467

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # tfrmt development version
 
+Improvements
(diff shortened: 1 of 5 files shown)


karima662 (approved):
looks good :)

PR #485, opened 2024-11-26
Update print_to_gt.R to put RTF footnotes in the document footer by default 

### LINKED ISSUE #422: put RTF footnotes in the document footer by default
Add this to the existing series of `tab_options` within the `print_to_gt` function: `tab_options(page.footer.use_tbl_notes = TRUE)`

This will get picked up if gt exports the table to an RTF document, but ignored otherwise.

alanahjonas95:
resolves #422

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,5 @@
 # tfrmt development version
 
-# tfrmt 0.1.4
 
 Bug fixes
 * Updated RTF footnotes so they go in the document footer by default.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # tfrmt development version
 
+# tfrmt 0.1.4
+
+Bug fixes
+* Updated RTF footnotes so they go in the document footer by default.
+
 # tfrmt 0.1.3
 
 Improvements
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # tfrmt development version
 
 
-Bug fixes
+Improvements
 * Updated RTF footnotes so they go in the document footer by default.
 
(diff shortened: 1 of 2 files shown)


bzkrouse:
@alanahjonas95 Thanks!! Can you please check this does what we want for saving to RTF, and doesn't impact other outputs? You could use {tfrmtbuilder} to make this quicker!

alanahjonas95:
@bzkrouse just made this quickly with tfrmtbuilder as suggested :)
looks like this with rtf now:
image
but all other methods appear with it just underneath eg:
this for png
image
and this for docx
image

bzkrouse:
@alanahjonas95 Perfect thank you!! One more thing, can you please add to the NEWS file ?

alanahjonas95:
> @alanahjonas95 Perfect thank you!! One more thing, can you please add to the NEWS file ? Then we should be good to go.

I've updated the read me but wasn't sure if i should mark as a bug fix or an improvement? and will the next release be 0.1.4?

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # tfrmt development version
 
+# tfrmt 0.1.4
```
bzkrouse, line 3:
```suggestion 
```

bzkrouse: we can just nest under "tfrmt development version" header for now :)

NEWS.md
```diff
@@ -1,5 +1,10 @@
...
+
+Bug fixes
```
bzkrouse, line 5:
```suggestion
Improvements
```

bzkrouse (approved):
Thank you!!

PR #487, opened 2024-12-05
quote handling in structures

### LINKED ISSUE #466: Single-quote escaping does not work for group_val/label_val arguments
Example

```
dd <- tibble( 
  label = c('val\'s', 'val a','val a'),
  value = c(1,2,3),
  col = c("a","a","a"),
  param = c("n","n","n")
)

tfrmt(label = label,
      value = value, 
      column = col,
      param = param,
      body_plan = body_plan(
        frmt_structure(group_val = '.default',
                       label_val = 'val\'s',
                       frmt("xx.xx"))
      )) |>
  print_to_gt(dd)
```
results in:

```
Error in `map()`:
ℹ In index: 1.
Caused by error in `parse()`:
! <text>:1:21: unexpected symbol
1: `label` %in% c('val's
                        ^
Run `rlang::last_trace()` to see where the error occurred.
```

bzkrouse:
resolves #466

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # tfrmt development version
 
+Bug fixes 
+* Fix issue where `*_structure` functions did not correctly parse strings containing quotes in some cases (#466)
+
 # tfrmt 0.1.3
 
 Improvements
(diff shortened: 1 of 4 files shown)


alanahjonas95 (approved):
looks good :)

PR #495, opened 2025-04-10
Added note for the frmt_combine

tcoroller:
Hello, 

This is a nice package, thanks for sharing!
It took me a while to understand the `frmt_combine`, especially since there is only one example with the efficacy.

I hope this comments can raise other users attention, and not spend time figuring out how to make it work.
Its not trivial the link between `param` and `label` 

Thanks again!

### DIFF
--- a/vignettes/examples.Rmd
+++ b/vignettes/examples.Rmd
@@ -545,6 +545,8 @@ For this example, we will use the `data_efficacy` dataset, an example analysis r
 
 This data is relatively simple in that it contains only 1 `group` column and `column` column, but it adds complexity in that multiple analyses are stacked together - summary statistics of different values at different time points, and the results of several different ANCOVA models. Multiple treatment groups as well as the contrasts between groups are included.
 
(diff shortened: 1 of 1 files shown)


vignettes/examples.Rmd
```diff
@@ -545,6 +545,8 @@ For this example, we will use the `data_efficacy` dataset, an example analysis r
 
 This data is relatively simple in that it contains only 1 `group` column and `column` column, but it adds complexity in that multiple analyses are stacked together - summary statistics of different values at different time points, and the results of several different ANCOVA models. Multiple treatment groups as well as the contrasts between groups are included.
 
+*NOTE:* Pay attention that `label` in the `data_efficacy` must be matching the `label_val` in the `label_val` for the `frmt_combine()` to work. For example, for `Mean (SD)`, each variable has differnt
```
bzkrouse, line 548:
```suggestion
*NOTE:* Pay attention that `label` in the `data_efficacy` must be matching the `label_val` in the `label_val` for the `frmt_combine()` to work. For example, for `Mean (SD)`, each variable has different 
```

PR #507, opened 2025-06-06
allow single level for page plan group

### LINKED ISSUE #506: page plan does not work if there is only 1 level in the group variable

shannonhaughton:
fix #506


bzkrouse (changes requested):
Looks great, thanks for getting this in so quickly!
can you please add a note to the NEWS file?

bzkrouse (approved):

PR #517, opened 2025-07-15
fix bug with big_n check not working, for sorting and for the error m…

### LINKED ISSUE #505: create a better error message for `check_big_n_page ` function
currently if the number of bigs supplied does not match the number of groups then the following error is given

```
Error in !all.equal(expected_grp_levs, actual_grp_levs, check.attributes = FALSE) : 
  invalid argument type
```

this Is for when someone wants to page by variables and has  ` by_page = TRUE` in the `big_n_struture`

needs updating to give a more informative error message 
https://github.com/GSK-Biostatistics/tfrmt/blob/b4cc81347c21e1c46474be9bdd03524608962758/R/apply_tfrmt.R#L489

### LINKED ISSUE #516: error in `check_big_n_page` when page plan variable sort order is changed

alanahjonas95:
…essage not appearing

resolves #516 
resolves #505

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,8 +5,9 @@ Improvements
 * Updated RTF footnotes so they go in the document footer by default.
 
 Bug fixes 
-* Fix issue where `*_structure` functions did not correctly parse strings containing quotes in some cases (#466)
-* Fix issue where page_plan doesn't work if there is only one level in your paging variable. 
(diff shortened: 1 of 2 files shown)


alanahjonas95:
> Looks great!!! can you add a couple of tests that capture the behavior? Should make sense to put them in test-big_ns.R

sure will do :)

bzkrouse (approved):

PR #522, opened 2025-07-21
Page plan bug

### LINKED ISSUE #497: big_n_structure function does not work when the column name has a space in it

alanahjonas95:
resolves #497 

allowing someone to put a space in the variable name used for grouping for the page plan when big n is also used


bzkrouse:
@alanahjonas95 looks good,  can you add to the test that the page notes are in correct order, for good measure. i.e. checking that result of this is what you would expect:

`map_chr(output_list, ~attr(.x, ".page_note"))`

bzkrouse (approved):

PR #529, opened 2025-08-29
remove prep_card, add vignette

bzkrouse:
* remove prep_card function (I commented out the tests in case some are still useful in the helpers...)
* Add a vignette
* add order_rows arg for shuffle_card() for compatibility with the cards hierarchical sorting utility

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,7 +24,7 @@ BugReports: https://github.com/GSK-Biostatistics/tfrmt/issues
 Depends: 
     R (>= 4.1.0)
 Imports: 
-    cards,
+    cards (>= 0.6.0),
     cli,
     dplyr,
     forcats,


DESCRIPTION
```diff
@@ -24,7 +24,7 @@ BugReports: https://github.com/GSK-Biostatistics/tfrmt/issues
 Depends: 
     R (>= 4.1.0)
 Imports: 
-    cards (>= 0.6.1.9010),
+    cards,
```
shannonhaughton, line 27: this needs some sort of minimum version because the functions don't run with the older versions of cards

shannonhaughton, line 27: this still needs a minimum version because shuffle_card doesn't work with cards 0.4.0

shannonhaughton: image

PR #542, opened 2025-09-23
Fix for upcoming stringr 1.6.0 release

hadley:
`str_replace()` no longer accepts `NA` patterns, which was occuring here because `1:length(x)` does not work well if x is length 0.


bzkrouse:
@hadley Thank you!

PR #545, opened 2025-09-24
Fix for gt v1.1.0

dragosmg:
{gt} v1.1.0 is already on CRAN and this PR solves the recently surfaced issues:

* bumps version number to dev
* solves #541 - errors with the upcoming gt release.
  * the errors are related to `tfrmt$label` either being empty or the column not being present in the data


dragosmg:
@bzkrouse On my machine I am getting 3 warnings regarding the tidyselect helpers not being used in a select context. These all surface from gt though. I'm going to mark with PR as ready for review.

PR #552, opened 2025-09-26
first pass for multi-column groups

### LINKED ISSUE #499: Ability to have multiple grouping variables in their own columns
Currently multiple grouping columns are indented and we have a max of 2 columns in the stub.
Add functionality to place these in separate columns instead.
until <issue> is implemented, we have to remove reliance on gt's stub

bzkrouse:
resolves #499


shannonhaughton:
I know this is draft but just reminder to add  a dep on gt version

Issue #553, opened 2025-09-29
add instructions for installing from CRAN to README


bzkrouse:
https://github.com/GSK-Biostatistics/tfrmt?tab=readme-ov-file#installation

PR #555, opened 2025-09-30
How install tfrmt from CRAN

JacobAyembilla:
#553  Instructions for installing tfrmt from CRAN


bzkrouse (commented):
Thank you @JacobAyembilla! We will take this and edit further!

PR #581, opened 2025-11-05
simplify replacement of unicode whitespace

### LINKED ISSUE #530: inside padding of cells is not unicode whitespace
Image

when using frmt_combine the whitespace within the string is not unicode whitespace which changes the alignment when it goes out to pdf

Image

p = frmt_when("==100" ~ "(100%)",
                      "==0" ~ "",
                      "<10" ~ frmt("( xx%)"),
                      TRUE ~ frmt("(xxx%)"))
can be resolved with:
p = frmt_when("==100" ~ "(100%)",
                      "==0" ~ "",
                      "<10" ~ frmt("(\U00A0xx%)"),
                      TRUE ~ frmt("(xxx%)"))

shannonhaughton:
fix #530

### DIFF
--- a/tests/testthat/test-print_to_gt.R
+++ b/tests/testthat/test-print_to_gt.R
@@ -15,6 +15,19 @@ test_that("convert_ws_unicode works as expected",{
 
   # function to apply
   whitespace_function <- gt_with_unicode$`_transforms`[[1]]$fn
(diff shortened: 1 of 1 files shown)


bzkrouse (changes requested):
looks great can you also add tests for there being multiple spaces in a row? like: 
value = "   (  x   x  )   "

bzkrouse (approved):

PR #620, opened 2026-01-30
update to remove lapply 

### LINKED ISSUE #572: use `fct_as_chr = TRUE` arg in `shuffle_card`
as opposed to this https://github.com/GSK-Biostatistics/tfrmt/blob/main/R/shuffle_card.R#L95

alanahjonas95:
resolves #572 
need to fix cards issue before this will work

cards issue fixed waiting for next release https://github.com/insightsengineering/cards/issues/542


tests/testthat/_snaps/shuffle_card.md
```diff
@@ -147,17 +147,19 @@
         cards::ADSL, variables = AGEGR1), 1), dplyr::slice(cards::ard_continuous(
         cards::ADSL, by = SEX, variables = AGE), 1), dplyr::slice(cards::ard_continuous(
         cards::ADSL, variables = AGE), 1)), by = c("ARM", "SEX")))
+    Message
+      Mismatch between attributes of `x` and supplied value to `by`. Attributes will be used in lieu of `by`
```
bzkrouse: @alanahjonas95 Did you look into this message? 

I think it has to do with this update - https://github.com/insightsengineering/cards/issues/483 
And what happens when you do bind_ard() on two different cards objects with differing by/variable/strata (spoiler: it keeps attributes from the first ARD in the sequence). Anyway I think it's a bug for cards. See this reprex:

```
 ard <- cards::bind_ard(
      cards::ard_categorical(cards::ADSL, by = ARM, variables = AGEGR1),
      cards::ard_continuous(cards::ADSL, by = SEX, variables = AGE)
    )

attributes(ard)
```

I think if we fix in cards then we don't need to code around it here. Can you please have a look?

R/shuffle_card.R
```diff
@@ -94,20 +94,18 @@ shuffle_card <- function(x,
...
-        ),
-        ~ lapply(., \(x) if (!is.null(x)) as.character(x))
-      ),
       stat_variable = .data$variable
     ) |>
+    cards::rename_ard_columns(
+      fill='..cards_overall..') |>
+    cards::unlist_ard_columns(
+      columns = c(cards::all_ard_groups(), cards::all_ard_variables()),
+      fct_as_chr = TRUE,
```
bzkrouse, line 103: does this require a newer version of cards?

alanahjonas95: yes it does, good point! i will update in description

bzkrouse (approved):

PR #628, opened 2026-03-09
create an extract data function

### LINKED ISSUE #493: add function to output the data behind the table
proposal for a feature to output the dataset underlying the tfrmt which can then be used for QC purposes. 

tfrmt(...) |>
print_to_gt(data) |>
extract_data()

alanahjonas95:
resolves #493

works on gt_tbl and gt_group outputs from print_to_gt
added a test for gt_tbl and gt_group outputs


bzkrouse (changes requested):
Nice!! Next step is to try this with a table with spanning headers/big Ns. I think we'll want to manipulate the data from the gt to incorporate this information in some way. It's not going to be pretty by default. 
Have a look at what it looks like and then we can chat :)

bzkrouse:
@alanahjonas95 the variable names in the big N example look fine as-is to me, since they have been formatted already. I don't love keeping the spanners with the "__tlang_delim__" so would prefer to swap that out. Maybe we could have an arg for extract_data that indicates which delimiter to use, i.e. extract_data(x, col_delim = "__"). What do you think? Can you add a few tests to demonstrate different levels of spanners?

alanahjonas95:
@bzkrouse 
Added col_delim argument that replaces "___tlang_delim___" with the specified delimeter, i wasnt sure if we want the extra underscores or not? can easily change that

So in this one set col_delim="_"
image

added couple tests for 1 and 2 layers of spanning headrs and one with a page plan
and other tests for the extract data still work without this argument/ doesnt do anything if this argument is used


added test to check that if the columns are renamed in the col plan than these renamed values do appear in the extracted data

R/extract_data.R
```diff
@@ -0,0 +1,42 @@
+#' Extract underlying data from tfrmt output
+#'
+#' Following a call to `print_to_gt`, this function extracts the underlying
+#' data frame(s) from the resulting `gt` or `gt_group` object.
+#'
+#' @param x A `gt_tbl` or `gt_group` object (usually the output of `print_to_gt()`).
+#' @param col_delim Character string to replace the internal "tlang_delim"
```
bzkrouse, line 57: I think we should have a delimiter default, can we do "_" by default?

tests/testthat/test-extract_data.R
```diff
@@ -0,0 +1,339 @@
...
+    ),
+    col_plan = col_plan(
+      group,
+      label,
+      starts_with("col")
+    )
+  )|> print_to_gt(data)
+
+
+  res_layer <- extract_data(spanning_tfrmt, col_delim="_")
```
bzkrouse: if you set col_delim to be "_" as i suggest above, let's make sure we add tests for setting a different col_dim

R/extract_data.R
```diff
@@ -0,0 +1,42 @@
...
+#'   separator in column names only for tables with spanning headers. Defaults to NULL (no replacement).
+#' @return If `gt_tbl`, a single data frame. If `gt_group`, a list of data frames.
+#' @importFrom purrr map
+#' @export
+extract_data <- function(x, col_delim = NULL) {
+
+  # helper to clean names
+  clean_names <- function(df, delim) {
```
bzkrouse: can you pls define clean_names outside of extract_data

R/extract_data.R
```diff
@@ -0,0 +1,42 @@
...
+#' Following a call to `print_to_gt`, this function extracts the underlying
+#' data frame(s) from the resulting `gt` or `gt_group` object.
+#'
+#' @param x A `gt_tbl` or `gt_group` object (usually the output of `print_to_gt()`).
+#' @param col_delim Character string to replace the internal "tlang_delim"
+#'   separator in column names only for tables with spanning headers. Defaults to NULL (no replacement).
+#' @return If `gt_tbl`, a single data frame. If `gt_group`, a list of data frames.
+#' @importFrom purrr map
+#' @export
+extract_data <- function(x, col_delim = NULL) {
```
bzkrouse: within this function, let's also drop any variables tfrmt adds under the hood, i.e. ..tfrmt_row_grp_lbl

bzkrouse (changes requested):
This is looking really great!!! 
Added a couple comments/suggestions in the PR.
Aside from those, there is one more piece to do related to column names. Your examples with col_plan renaming show that the regular data columns have the renamed value instead of the original. This is not the case for stub columns (try to rename group/label in the col_plan of your examples and you'll see that). The stub labels are stored within the gt object (specifically x$`_stubhead`) and applied when the gt is built (similar to how the spanners are arranged when the gt is built). i think we should either have everything relabeled or nothing relabeled - since you already have the other columns relabeled, can you try to also relabel the stub columns? 
For a test, you'll need to set row_grp_plan label loc to "column" to have multiple stub cols. 

Thanks!

R/extract_data.R
alanahjonas95: sure updated

tests/testthat/test-extract_data.R
alanahjonas95: added test for setting col_delim = "/"

R/extract_data.R
alanahjonas95: yes moved out

R/extract_data.R
alanahjonas95: yeah good idea! added this to the clean data little function so can do in one place

alanahjonas95:
> This is looking really great!!! Added a couple comments/suggestions in the PR. Aside from those, there is one more piece to do related to column names. Your examples with col_plan renaming show that the regular data columns have the renamed value instead of the original. This is not the case for stub columns (try to rename group/label in the col_plan of your examples and you'll see that). The stub labels are stored within the gt object (specifically x$`_stubhead`) and applied when the gt is built (similar to how the spanners are arranged when the gt is built). i think we should either have everything relabeled or nothing relabeled - since you already have the other columns relabeled, can you try to also relabel the stub columns? For a test, you'll need to set row_grp_plan label loc to "column" to have multiple stub cols.
> 
> Thanks!

added functinality to extract/rename the stub_head names from group/label columns if they have been renamed

alanahjonas95:
@bzkrouse not quite sure why one of the test is failing im not sure it relates to my code?

R/extract_data.R
```diff
@@ -46,11 +87,15 @@ extract_data <- function(x, col_delim = "_") {
...
+                                                 delim = col_delim,
+                                                 boxhead = .x[["_boxhead"]],
+                                                 stubhead = .x[["_stubhead"]]
+    ))
 
     return(extracted_list)
   }
 
-  # Fallback
   stop("Input must be a 'gt_tbl' or 'gt_group' object.")
```
bzkrouse: do this checking up front :)

R/extract_data.R
```diff
@@ -4,12 +4,48 @@
...
+
+    if (length(stub_vars) > 0 && length(stub_vars) == length(new_stub_labels)) {
+
+      lookup <- setNames(stub_vars, new_stub_labels)
+
+      #  Filter out entries where the new name is empty or NA
+      valid_names <- names(lookup) != "" & !is.na(names(lookup))
+      lookup <- lookup[valid_names]
+
+      #Only keep names where the new name is DIFFERENT from the old name
```
bzkrouse: does this line and the next line matter?

bzkrouse (changes requested):
this looks really great! I put a couple Qs/comments in the PR. 
Testing this, the extracted data doesn't necessarily reflect the order it appears in the table, or the data that goes into tfrmt - see the position of rowlbl1 in the result of your first test. Can you investigate?

R/extract_data.R
alanahjonas95: moved this check up

R/extract_data.R
alanahjonas95: no i guess not will remove it as wont make any difference

alanahjonas95:
> this looks really great! I put a couple Qs/comments in the PR. Testing this, the extracted data doesn't necessarily reflect the order it appears in the table, or the data that goes into tfrmt - see the position of rowlbl1 in the result of your first test. Can you investigate?

im a bit confused, rowlb1 isn't in the extracted data it just shows rowlblb2 with both values in the one column, for the first test and i think they are in the order of the table? 
image
image

bzkrouse:
@alanahjonas95 You are right, I have no idea why I was seeing something different at the time :) Did one more pass through and it looks great, can you please do one more update to use pkg::fun() for the function calls, instead of importfrom pkg fun (it's what we want to move towards, you currently have a mix). 
Also, please add a NEWS file and a small vignette.
Thank you!

alanahjonas95:
@bzkrouse ive made the suggested updates and created a vignette but i am unsure how to test render it?

alanahjonas95:
> @bzkrouse ive made the suggested updates and created a vignette but i am unsure how to test render it?

@bzkrouse 
checked now thanks :) let me know what you think

bzkrouse (changes requested):
looks fabulous!
So, since you did this, we learned about this function - https://gt.rstudio.com/reference/extract_body.html

Could you please have a look at whether this should be used to properly pull the data from the gt object? Any benefits or unwanted side effects?

alanahjonas95:
> looks fabulous! So, since you did this, we learned about this function - https://gt.rstudio.com/reference/extract_body.html
> 
> Could you please have a look at whether this should be used to properly pull the data from the gt object? Any benefits or unwanted side effects?

so thought i would try the new gt function on some of the tests ive written so we can compare the outputs? probably the easiest way?

So this is what my function vs the gt one give for simple example:
image
so there are a few differences, 

Column names

- extract_data() (current): rowlbl2
- extract_body() (gt): ::rowname:: — the internal gt stub name, would need renaming

Group header rows (Age (y), Sex)

- extract_data(): NA across all value columns
- extract_body():  `<br />` in all value columns — gt has injected HTML line break tags during its build pipeline

HTML encoding in values

- extract_data(): <65 yrs, >80 yrs (raw characters)
- extract_body(): `&lt;65 yrs,`  `&gt;80 yrs` — gt has HTML-escaped the `</>` characters

Value formatting

- extract_data(): " 86" (with leading space, quoted)
- extract_body(): 86 (unquoted, no leading space) — gt has stripped the tfrmt formatting whitespace during rendering

alanahjonas95:
for my second test where the labels are updated im getting the following error from the gt function:
> tfrmt_data_extracted <- extract_data(tfrmt_dem)
>  tfrmt_data_gt <- gt::extract_body(tfrmt_dem)
Error in if (is.na(rowname_col)) { : the condition has length > 1


image

and doesnt seem to work on a gt_group object:

image

alanahjonas95:
works pretty much same as first mentioned difference for table with bign values:
image

alanahjonas95:
for spaning headers works but doesnt remove __tlang_delim__
image

alanahjonas95:
differences summary:

| Feature / Concern | Your `extract_data()` | `gt::extract_body()` |
| :--- | :--- | :--- |
| **Data source** | `[["_data"]]` (raw input) | `[["body"]]` (processed, formatted strings) |
| **Column filtering** | Manual: drops `..tfrmt*`, renames via boxhead/stubhead | Uses gt's own `dt_boxhead_get_vars_default()` — properly excludes hidden cols |
| **Stub/group col names** | Reads from `_stubhead`/`_boxhead` labels | Renames to `"::rowname::"` / `"::group_id::"` (not useful for you) |
| **Spanning header names** | Your `tlang_delim` &rarr; `col_delim` replacement | Doesn't change spanning headers — returns internal column names from `_body` |
| **`gt_group` support** | ✅ You handle it | ❌ Only accepts `gt_tbl` |
| **`bigN` in col names** | ✅ You pull from boxhead labels | ❌ Returns original variable names, not the formatted label strings |

bzkrouse:
thanks @alanahjonas95 ! I was wondering if using extract_body is better than gt$`tbl` (you'd still need your extra manipulations. I am concerned about the extra html characters that are added though. Shall we keep the way you have i then?

alanahjonas95:
> thanks @alanahjonas95 ! I was wondering if using extract_body is better than gt$`tbl` (you'd still need your extra manipulations. I am concerned about the extra html characters that are added though. Shall we keep the way you have i then?

happy to keep as it is if you are? is there anything else you want me to look into?

bzkrouse (approved):

Issue #518, opened 2025-07-16
warning message when grouping variable is a factor


alanahjonas95:
getting the following warning message when our grouping variable siteid is a factor

```
Unable to apply row group structure
Reason: In argument: `across(...)`.
```

example tfrmt code

```
tfrmt_n_pct(n = "n", pct = "p") |>
  tfrmt(
    title = "Table AE_T01: Summary of Adverse Events by System Organ Class and Preferred Term",
    group = c(SITEID, AEBODSYS),
    label = AETERM,
    param = stat_name,
    value = stat,
    column = TRT01A,
    col_plan = col_plan("System Organ Class
            Preferred Term" = AEBODSYS, -ord1,  -ord2,-SITEID),
    page_plan = page_plan(
      page_structure(
        group_val = list(SITEID = ".default")),
      note_loc = "subtitle"
    ) ,
    row_grp_plan = row_grp_plan(row_grp_structure(
      group_val = ".default", element_block(post_space = " ")
    )),
    big_n = big_n_structure(param_val = "bigN", n_frmt = frmt("\n(N=xx)"), by_page = TRUE)
  ) |>
  print_to_gt(ae5_ard |>
                mutate(SITEID=factor(SITEID, levels = c("703 (N=18)", "704 (N=25)", "701 (N=41)")))|>
                arrange(SITEID, ord1, ord2))


```

bzkrouse:
@alanahjonas95 I took a closer look and I think this one needs some more thought. The row_grp_plan is trying to add rows of characters (blank spaces in this case). This will work fine for character columns. In order for this to successfully work on factor columns as well, we need to first coerce them to character. This is a pretty big change so I am inclined to take care of it upstream in the data process. 
We also need to understand how coercing to character may impact other bits of formatting. For example, what if someone passes a factor grouping column to both the `group` and `sorting_cols` parameter of tfrmt? (should we allow that at all)?

PR #629, opened 2026-03-11
fix Conflict between row_grp_plan and page_plan max_rows

### LINKED ISSUE #539: Conflict between row_grp_plan and page_plan max_rows
Seems to occur if a grp value is an empty string like so:

```
df <- tibble::tribble(
  ~ grp, ~ lbl, ~ prm, ~ trt,
  ""  ,  "summ" , "n"  , 22,
  "A"  ,  "a" , "n"  , 22,
  "A"  ,  "b" , "n"  , 11,
  "B"  ,  "a" , "n"  , 24,
  "B"  ,  "b" , "n"  , 55
) %>%
  pivot_longer(trt, names_to = "column", values_to = "value")

mytfrmt <- tfrmt(
  group = "grp",
  label = "lbl",
  param = "prm",
  column = "column",
  value = "value",
  body_plan = body_plan(frmt_structure(group_val=".default", label_val=".default", frmt("xx"))),
  row_grp_plan = row_grp_plan(
    row_grp_structure(group_val = ".default", element_block(post_space = " "))
  ),  
  page_plan = page_plan(max_rows = 4)
) 
print_to_gt(mytfrmt, df)
```

returns this error:
```
Unable to to apply apply_page_plan.
Reason: In index: 1.
```

alanahjonas95:
resolves #539 

ive updated the apply_page_max_rows function to convert group values blanks to a space to fix the issue

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 ## Bug fixes
 * Fix bug in `tfrmt_sigdig` so it correctly passes the 'missing' argument to the body_plan (#621, @alanahjonas95).
+* Fix bug in `page_plan` where the max_rows argument was returning an error if a value in the group variable was an empty string (#629, @alanahjonas95).
 
 ## Improvements
 * `footnote_plan()` receives a new argument, `order`, allowing users to specify the order of footnotes (#605, @alanahjonas95).
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## Bug fixes
 * Fix bug in `tfrmt_sigdig` so it correctly passes the 'missing' argument to the body_plan (#621, @alanahjonas95).
-* Fix bug in `page_plan` where the max_rows argument was returning an error if a value in the group variable was an empty string (#629, @alanahjonas95).
+* Fix bug in `page_plan` where the max_rows argument was returning an error if a value in the group variable was an empty string (#539, @alanahjonas95).
 
 ## Improvements
 * `footnote_plan()` receives a new argument, `order`, allowing users to specify the order of footnotes (#605, @alanahjonas95).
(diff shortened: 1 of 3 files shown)


R/apply_page_plan.R
```diff
@@ -54,7 +54,12 @@ apply_page_plan <- function(.data, page_plan, group, label, row_grp_plan_label_l
 #' @importFrom forcats fct_inorder
 apply_page_max_rows <- function(.data, max_rows, group, label, row_grp_plan_label_loc){
 
+  # Ensure empty strings in grouping columns are treated as literal spaces
+  # so pagination math doesn't break
+  group_cols <- map_chr(group, rlang::as_label)
+
   .data <- .data %>%
+    mutate(across(all_of(group_cols), ~if_else(.x == "", " ", .x))) %>%
```
bzkrouse: can you add a condition here to only apply this transformation on character variables, not factors?

fyi because of #518, row group plan cant successfully add blank rows when it's a factor variable. So the likelihood of factor + "" is low anyway.

alanahjonas95: added a check to only apply the transformation to character group variables

added a test but removed the row group plan as that causes a note the one in the issue you linked, but the page plan does work when there is a blank value when the column is a factor without any transformation

bzkrouse (commented):
looks good! can you add a NEWS entry please :)

NEWS.md
```diff
@@ -2,6 +2,7 @@
 
 ## Bug fixes
 * Fix bug in `tfrmt_sigdig` so it correctly passes the 'missing' argument to the body_plan (#621, @alanahjonas95).
+* Fix bug in `page_plan` where the max_rows argument was returning an error if a value in the group variable was an empty string (#629, @alanahjonas95).
```
bzkrouse:
```suggestion
* Fix bug in `page_plan` where the max_rows argument was returning an error if a value in the group variable was an empty string (#539, @alanahjonas95).
```

bzkrouse (approved):

PR #630, opened 2026-03-11
add option to row_grp_plan/row_grp_structure to not add a space after the last group

### LINKED ISSUE #604: add option to row_grp_plan/row_grp_structure to not add a space after the last group
so not to have a space at the end of the table currently have to list out row_grp_structure for each group if you dont want a blank row at the end.

alanahjonas95:
resolves #604


R/apply_tfrmt.R
```diff
@@ -175,6 +175,21 @@ apply_tfrmt_subtable <- function(tfrmt, .data, col_plan_vars, stub_header, big_n
       tfrmt$label
     )
 
+  # Check if the tfrmt_post_space_row column exists AND if the last value is TRUE
```
bzkrouse: can you wrap this piece into a function please :)

alanahjonas95: added a function for this let me know what you think :)

R/apply_row_grp_plan.R
```diff
@@ -87,9 +87,12 @@ apply_row_grp_struct <- function(.data, row_grp_struct_list, group, label = NULL
                           }
                         }) %>%
     arrange(.data$TEMP_row) %>%
+    mutate(..tfrmt_post_space_row = case_when(TEMP_row %% 1 != 0 ~ TRUE,
```
bzkrouse: The apply_row_grp_block is a general utility and not necessarily specific to post spaces. Have a look at the apply_grp_block function as it contains logic specific to the post spaces, to see if you can incorporate there instead.

R/apply_tfrmt.R
```diff
@@ -143,10 +164,10 @@ apply_tfrmt_subtable <- function(tfrmt, .data, col_plan_vars, stub_header, big_n
       fail_desc = "Unable to add big N's"
     )
 
-
   tbl_dat_wide_processed <- .data  %>%
     #Select before grouping to not have to deal with if it indents or not
-    tentative_process(apply_col_plan, col_plan_vars,
+    tentative_process(apply_col_plan,
+                      append(col_plan_vars, rlang::quos(tidyselect::any_of("..tfrmt_post_space_row"))),
```
bzkrouse, line 149: why was this needed? is there a specific test that captures why this was required?

R/apply_row_grp_plan.R
alanahjonas95: moved this

R/apply_tfrmt.R
alanahjonas95: so its needed here as the `..tfrmt_post_space_row` column is added in `apply_grp_block` but `apply_col_plan` subsets columns to only those specified in the user's `col_plan` — so the column would be silently dropped before `apply_post_space_trim` ever gets to use it.

alanahjonas95: have added a new test to confirm the ..tfrmt_post_space_row is dropped

R/apply_tfrmt.R
```diff
@@ -120,6 +120,27 @@ apply_tfrmt_subtable_mapper <- function(tfrmt, .data, col_plan_vars, stub_header
   }
 }
 
+#' Remove trailing post-space rows and helper column
+#' @param .data processed wide tbl
+#' @noRd
+apply_post_space_trim <- function(.data) {
```
bzkrouse: I think this function belongs in the apply_row_grp_plan file with those utilities

bzkrouse (changes requested):
Looking great! just one more thing :)

R/apply_tfrmt.R
alanahjonas95: sure ive moved it let me know if this is ok :)

bzkrouse:
@alanahjonas95  totally forgot - can you add a NEWS entry pls?

alanahjonas95:
> @alanahjonas95 totally forgot - can you add a NEWS entry pls?

added entry to news file

NEWS.md
```diff
@@ -7,6 +7,7 @@
 ## Improvements
 * `footnote_plan()` receives a new argument, `order`, allowing users to specify the order of footnotes (#605, @alanahjonas95).
 * Add markdown processing of stub column labels (#617)
+* Add option to `row_grp_plan()`/`row_grp_structure()` to not add a blank row after the last group (#630, @alanahjonas95).
```
bzkrouse: @alanahjonas95 This isn't an option though, right, it's just default behavior? Can you please update wording so it is clear what the behavior is (also keep in mind its not always a blank row of spaces, could be other characters - https://gsk-biostatistics.github.io/tfrmt/articles/row_grp_plan.html#element-block-objects)

alanahjonas95: updated hows it now

bzkrouse (approved):

PR #634, opened 2026-04-24
Fix bug when labels are repeated in different groups and changes the order of output

### LINKED ISSUE #633: Bug Report: If same label repeated they are all forced to be printed at the top
### What happened?

In this example it is desired to have an n row above header but they are being forced to be printed at the top, tried to use noprintvar but this doesnt work as then we lose the grouping header we do want in this examle we have two grouping vars for different parts of the table
for a simpler example i was able to use noprintvar and get the desired ordering for the n or we can remove group1 from group and outputs as desired for a simple table of n followed by np rows repeated same format

however with the following example with two different sections the n np rows and some summary stats at the bottom, 

```
library(tfrmt)
library(dplyr)

groups <- c("Group 5", "Group 6", "Group 7")

trt_a <- bind_rows(
  tibble(
    grouping  = rep(groups, each = 3),
    stat_name = rep(c("n", "n", "p"), times = 3),
    group1 = c(
      "n", "Hepatocellular injury", "Hepatocellular injury",
      "n", "Hys Law", "Hys Law",
      "n", "Total BIL", "Total BIL"),
    PARAM     = group1,
    stat      = c(154, 2, 1.298, 154, 1, 0.649, 154, 4, 2.597)
  ),
  tibble(
    grouping  = "Group 8",
    stat_name = c("n", "Mean", "SD", "Median"),
    group1    = "Time from First Dose",
    PARAM     = stat_name,
    stat      = c(7, 135.714, 120.227, 106)
  )
) |> 
  mutate(trt01a = "Drug A")

pbo_mults <- c(rep(c(1.0, 0.5, 0.5), 3), 0.8, 1.1, 0.9, 1.05)

trt_pbo <- trt_a |>
  mutate(
    trt01a = "Placebo",
    stat   = stat * pbo_mults
  )

df_final <- bind_rows(trt_a, trt_pbo)

tfrmt(
  group = c("group1", "grouping"),
  label = PARAM,
  param = stat_name,
  value = stat,
  column = trt01a,
  col_plan = col_plan(PARAM, -grouping),
  body_plan = body_plan(
    frmt_structure(group_val = ".default", label_val = ".default", frmt("xxx")),
    frmt_structure(
      group_val = ".default",
      label_val = ".default",
      frmt_combine("{n}{p}", n = frmt("xxx"), p = frmt(" (xx%)"))
    )
  )
) |>
  print_to_gt(df_final)

```

output looks like this:

Image

but desired to have the n row above each catgeory they relate to but they are all getting forced to the top as the labels are the same
with spaces:

Image

if i add a space / 2 spaces to the end of n values to make them distinct the issue does away and i can print them in the desired order

note if the table is just the top section then we can just drop group1 from group and the table will output as expected without adding the spaces, but with the  summary stats section dropping group1 results in the "time from first dose ..." label not printing

Image

alanahjonas95:
resolves #633


R/apply_row_grp_plan.R
```diff
@@ -280,8 +280,9 @@ combine_group_cols <- function(.data, group, label, element_row_grp_loc = NULL){
   while(length(group) > 0 & !is.null(label)){
 
     split_dat <- .data %>%
-      group_by(!!!top_grouping) %>%
-      group_split()
+      group_by(run_id = dplyr::consecutive_id(!!!top_grouping)) %>%
```
bzkrouse, line 281: Your reprex from the issue nicely demonstrates this bug. I took a look myself and observed that this logic is coming into play after the col plan has been applied, and our little dummy group variable is dropped. therefore we no longer have it here to dictate the group_by/splitting. 

Note that a few lines up in this function, we are trying to use fct_inorder to set the groups before the splitting occurs. Since our dummy group is gone, this is kind of pointless in this scenario. Does your code eliminate the need for fct_inorder? Is the data always sorted properly at this point in the apply_tfrmt logic?

alanahjonas95: Im not sure where this is happening struggling to find it? " and our little dummy group variable is dropped" could you link me to it? looked in apply_tfrmt but why i run the apply_col_plan  before/after i cant see a dummy group variable?

bzkrouse: sorry Alanah, poor description :) In the reprex there is an extra grouping variable that is used to organize the rows but then is dropped in the col_plan. I am referring to that extra grouping variable as the "little dummy group variable"

alanahjonas95: The ordering is wrong even if i keep that variable? 
image

alanahjonas95: if i remove the fct_inorder some of the tests break so seems its still needed? and a couple of tests were missing sorting columns which led to the tests failing with this changed logic so had to update that?

alanahjonas95: without the fct_inorder (line 263-265) of apply_row_grp_plan code, many tests in apply_row_grp_plan fail: 
image

bzkrouse: I think you are dropping the creation of the ..tfrmt_row_grp_lbl, but you should keep that and only try dropping the fct_inorder piece. They are both happening in the same mutate.

alanahjonas95: yeah you are right! ive updated to just remove the fct_inorder part and all the tests pass now with my updated logic

bzkrouse (changes requested):
can you add an update to the NEWS file :)

alanahjonas95:
@bzkrouse update added to news file

bzkrouse (approved):

PR #636, opened 2026-04-28
remove min version dep for stringr

### LINKED ISSUE #635: remove strict requirement on stringr dep
the code to replace spaces with unicode whitespaces requires a new version of stringr. 
https://github.com/GSK-Biostatistics/tfrmt/blob/main/R/print_to_gt.R#L504
Consider looking at revising the code to eliminate that strict dependency

bzkrouse:
closes #635 

@shannonhaughton I've tested your code for the convert_unicode_ws() with 1.5.2 and 1.6.0 and it works with both. Can you please confirm in case I missed something?

### DIFF
--- a/R/apply_col_plan.R
+++ b/R/apply_col_plan.R
@@ -345,7 +345,7 @@ split_data_names_to_df <- function(data_names, preselected_cols, column_names){
   }
 
   tibble(
-    original = data_names,
+    original = unname(data_names),
     new_name = names(data_names)
    ) %>%
     mutate(
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -37,7 +37,7 @@ Imports:
     purrr,
     rlang,
     stringi,
-    stringr (>= 1.6.0),
+    stringr,
     tibble,
     tidyr,
     tidyselect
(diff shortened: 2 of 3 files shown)


shannonhaughton:
@bzkrouse the breaking change was in behaviour of str_replace_all 
image
- I remember having different behaviour in the version I had on my local compared to the version the cmd checks were running with - looking into it

shannonhaughton:
I think the for loop in break_duplicate_whitespace means that it will work for both - I think you're good to remove it

shannonhaughton (approved):

R/apply_col_plan.R
```diff
@@ -345,7 +345,7 @@ split_data_names_to_df <- function(data_names, preselected_cols, column_names){
   }
 
   tibble(
-    original = data_names,
+    original = unname(data_names),
```
shannonhaughton, line 348: I'm not sure what this change is for, assuming something unrelated

bzkrouse: The new version of stringr preserves names from named vectors after str_* operations. So one of our tests was having different behavior in different versions. Figured it was easiest and cleanest to just strip the names here (although with more time could avoid adding the names in the first place..)

PR #641, opened 2026-06-24
add code to remove attributes if bind_ard

### LINKED ISSUE #640: Support bind_ard class from cards::bind_ard() by dropping top-level attributes in shuffle_card()
Following the resolution and approach discussed in [cards#572](https://github.com/insightsengineering/cards/pull/572), `cards::bind_ard()` will now assign a specific class to bound datasets (e.g., `bind_ard`), rather than letting it strictly retain the inaccurate top-level attributes of only the first ARD in the binding sequence.

Because `tfrmt::shuffle_card()` currently relies heavily on the by, variable, and strata attributes attached to an ARD, encountering a multi-ARD dataset with a bind_ard structure causes downstream evaluation failures, warnings, or incorrect column drops (such as dropping totals or failing to correctly resolve fill_overall).

We need to update `tfrmt::shuffle_card()` to recognise this new approach and intentionally drop those top-level attributes so it falls back to a clean, structural evaluation.

### Proposed Changes:

1. Class Check & Attribute Removal:

   - Update shuffle_card() to check if the incoming ARD inherits from the bind_ard class.
   - If it does, explicitly drop/strip the traditional by, variable, and strata top-level attributes from the object.

2. Structural Evaluation Fallback:

   - By removing these attributes, ensure shuffle_card() cleanly defaults to evaluating the data structurally row-by-row (or section-by-section) based on the actual columns and values present in the dataset.

alanahjonas95:
resolves #640

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -52,6 +52,8 @@ Suggests:
     survival,
     testthat (>= 3.0.0),
     withr
+Remotes: 
+    insightsengineering/cards
 Config/testthat/edition: 3
 Encoding: UTF-8
 Language: en-GB
--- a/R/shuffle_card.R
+++ b/R/shuffle_card.R
@@ -62,6 +62,13 @@ shuffle_card <- function(x,
     )
   }
 
+  # If a combined ARD is passed, drop stale attributes and evaluate structurally
+  if (inherits(x, "bind_ard") && !is.null(attr(x, "args"))) {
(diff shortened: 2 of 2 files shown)


alanahjonas95:
@bzkrouse tests are all passing when i use my updated version of cards
image

and here shuffle_card tests passing specifically:
image

bzkrouse (approved):

PR #643, opened 2026-06-25
Closes #642: Apply fix for column headers containing parenthesis

### LINKED ISSUE #642: Bug Report: `json_to_tfrmt` fails when reading in columns with brackets in the header
### **Description**
When running `json_to_tfrmt`, I am encountering an error when trying to read back in a column structure that contains a span over two columns when one of the child columns contains brackets (for example is formatted as `n (%)`.

Image

**Program flow:** `json_to_tfrmt` -> `ls_to_col_plan` -> `ls_to_span_structure` ->`parse_expr`

When the `span_ls` is created inside of `ls_to_span_structure` it looks like this with "n (%)" correctly wrapped in `""`:

Image

This gets passed back in the return after being turned into a `span_structure`. Then `as.character |> parse_expr` triggers the failure:

Image

This is the `span_structure`: 

Image

And finally the resulting string from `as.character(struct)` where you can see "n (%)" not wrapped in `''`:

Image

### **Reprex**
In the below .zip I have included the input data, the tfrmt object saved as a json file, a small script to run and encounter the bug, and pdf and rds files of the expected output generated with dummy data.#

[tfrmt_bug.zip](https://github.com/user-attachments/files/29312007/tfrmt_bug.zip)

### **SessionInfo**

Image

LiamHobby:
Closes #642 

**Summary of changes**
`json_to_tfrmt` threw a parse error when reading a `col_plan` that contained `span_structure` elements whose column names contained parenthesis i.e., "n (%)".

`as.character.span_structure()` converts a `span_structure` to a character string so it can be converted to JSON. Plain column names are quoted `"evt" -> 'evt'` but function calls should not be quoted. The existing regex `\\(.+\\) captured any string containing `(something)`.

"n (%)" contains "(%)" so it matched the regex and was incorrectly treated as a function call which then fails to evaluate in `parse_expr()`.

The new regex matches only strings that start with a valid R function identifier followed immediately by a parenthesis (no space between the name and the `(`.

New test added for round trip validation `tfrmt` -> `json` -> `tfrmt`.

**Note for the reviewer**
Running `devtools::test()` causes three snapshot failures in the file `test-shuffle_card.R` using `cards` version 0.8.0.
image

Downgrading to version 0.7.1 passes all tests:
image

**R CMD Check passes**
image

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,8 @@
 ## Bug fixes
 * Fix bug in `tfrmt_sigdig` so it correctly passes the 'missing' argument to the body_plan (#621, @alanahjonas95).
 * Fix bug in `page_plan` where the max_rows argument was returning an error if a value in the group variable was an empty string (#539, @alanahjonas95).
-
+* Fix bug in `as.character.span_structure` where the regex was incorrectly capturing content wrapped in parentheses as an R function even though it did not have a valid R function identifier immediately before the `(`. i.e., "n (%)" incorrectly captured as a function. (#643)
+  
 ## Improvements
 * `footnote_plan()` receives a new argument, `order`, allowing users to specify the order of footnotes (#605, @alanahjonas95).
 * Add markdown processing of stub column labels (#617)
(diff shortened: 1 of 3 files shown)


alanahjonas95 (approved):
Looks great thanks for the Fix and test @LiamHobby 
just waiting for a fix in cards then tests should all run succesfully

PR #650, opened 2026-07-06
fix bug in shuffle card preventing by attribute  being passed after  …

### LINKED ISSUE #651: shuffle_card handling of attributes
Couple of issues now that cards passes along attributes vs all the ard_* functions:

(1) cards::bind_ard has unreliable attributes due to the underlying dplyr::bind_rows just keeping attributes from the first object. Already decided on a fix to tfrmt::shuffle_card to drop the attributes of objects of class `bind_ard` (newly added class).

(2) For shuffle_card, should we prefer the `by` over the attributes?
Current logic prefers the attributes:
https://github.com/GSK-Biostatistics/tfrmt/blob/main/R/shuffle_card.R#L200
Given our pickle with `bind_ard` attributes, and the inability for user to override, rethinking this logic. 

In both (1) and (2) let's have clear messages to the user about what's going on. For (1) we should flag if object is class `bind_ard` and user has not provided a `by`. For (2) if we change behavior we should add a note to the "Mismatch in..." message about behavior being different as of tfrmt vx.x.x

alanahjonas95:
…bind_ard

resolves #651 
I've updated to add a cli abort message if the attributes has class bind_ard and there is no by variable supplied then we have a message to tell the user to supply the by variables

issue related to this: https://github.com/GSK-Biostatistics/tfrmt/issues/640

so currently ive done the following:
basically the options are:

if class bind_ard then drop attributes, message user if they have not supplied a by

but i could update to do this:
 use attributes of the input and message user what it's inferring for by

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -52,8 +52,6 @@ Suggests:
     survival,
     testthat (>= 3.0.0),
     withr
-Remotes: 
-    insightsengineering/cards
 Config/testthat/edition: 3
 Encoding: UTF-8
 Language: en-GB
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,7 +42,7 @@ Imports:
     tidyr,
     tidyselect
 Suggests:
-    cards (>= 0.8.1),
+    cards (>= 0.8.0),
     covr,
     ggfortify,
     knitr,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,7 +42,7 @@ Imports:
     tidyr,
     tidyselect
 Suggests:
-    cards (>= 0.8.0),
+    cards (>= 0.8.1),
     covr,
     ggfortify,
     knitr,
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,7 @@
 * Add markdown processing of stub column labels (#617)
 * Update default for `row_grp_plan()`/`row_grp_structure()` to remove trailing post-space rows (#630, @alanahjonas95).
 * Added `extract_data()` function to enable data to be easily extracted from a tfrmt into a data frame.(#628, @alanahjonas95)
-
+* `shuffle_card()` now explicitly handles `bind_ard` attributes, prioritising the user-supplied `by` argument over inherited ARD attributes and providing clearer warning messages during mismatches (#650, @alanahjonas95).
 
 # tfrmt 0.3.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,8 @@
 * Add markdown processing of stub column labels (#617)
 * Update default for `row_grp_plan()`/`row_grp_structure()` to remove trailing post-space rows (#630, @alanahjonas95).
 * Added `extract_data()` function to enable data to be easily extracted from a tfrmt into a data frame.(#628, @alanahjonas95)
-* `shuffle_card()` now explicitly handles `bind_ard` attributes, prioritising the user-supplied `by` argument over inherited ARD attributes and providing clearer warning messages during mismatches (#650, @alanahjonas95).
+* `shuffle_card()` now automatically strips inherited ARD attributes from objects of class `bind_ard` to avoid unreliable metadata (#650).
+* `shuffle_card()` now prioritises the user-supplied `by` argument over inherited ARD attributes, giving users control to override attributes and updating the mismatch message to reflect this change (#650).
 
 # tfrmt 0.3.0
 
--- a/R/shuffle_card.R
+++ b/R/shuffle_card.R
@@ -64,11 +64,13 @@ shuffle_card <- function(x,
 
   # If a combined ARD is passed, drop stale attributes and evaluate structurally
   if (inherits(x, "bind_ard") && !is.null(attr(x, "args"))) {
+    if (!is.null(by)) {
       attr(x, "args")$by <- NULL
+    }
       attr(x, "args")$variable <- NULL
       attr(x, "args")$strata <- NULL
   }
-
+  
   ard_args <- attributes(x)$args
   by <- .process_by(x, by)
 
--- a/R/shuffle_card.R
+++ b/R/shuffle_card.R
@@ -200,7 +200,7 @@ shuffle_card <- function(x,
       cli::cli_inform(
         c("Mismatch between attributes of {.arg x} and supplied value to {.arg by}.",
           "i" = "Supplied value will be used in lieu of attributes.",
-          "*" = "Note: As of {.pkg tfrmt} v0.4.0, explicitly supplied {.arg by} values take priority over data frame attributes."),
+          "*" = "Note: As of {.pkg tfrmt} v0.4.0, explicitly supplied {.arg by} values take priority over the {.cls card} attributes."),
         env = rlang::caller_env()
       )
     }
--- a/R/shuffle_card.R
+++ b/R/shuffle_card.R
@@ -72,6 +72,13 @@ shuffle_card <- function(x,
   ard_args <- attributes(x)$args
   by <- .process_by(x, by)
 
+  # Check if a 'by' variable is available
+  if (is.null(by) || length(by) == 0) {
+    cli::cli_abort(
+       "The {.arg by} argument was not supplied."
+      )
+  }
+
   # make sure columns are in order & add index for retaining order
   if (isTRUE(order_rows)){
     x <- x |>
--- a/R/shuffle_card.R
+++ b/R/shuffle_card.R
@@ -391,7 +391,7 @@ shuffle_card <- function(x,
   output <- dplyr::case_when(
     x == "..cards_overall.." ~ overall_val,
     x == "..hierarchical_overall.." ~ any_val,
-    TRUE ~ x
+    TRUE ~ as.character(x)
   )
 
   output
--- a/R/shuffle_card.R
+++ b/R/shuffle_card.R
@@ -189,7 +189,7 @@ shuffle_card <- function(x,
 #' Process `by` variable
 #'
 #' @param x a data frame
-#' @param by Grouping variable(s) used in calculations. Defaults to `NULL`.
+#' @param by Grouping variable(s) used in calculations. 
 #'
 #' @returns character string if `by` variable present
 #' @noRd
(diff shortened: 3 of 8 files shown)


R/shuffle_card.R
```diff
@@ -72,6 +72,16 @@ shuffle_card <- function(x,
   ard_args <- attributes(x)$args
   by <- .process_by(x, by)
 
+  # Check if a 'by' variable is available
+  if (inherits(x, "bind_ard") && (is.null(by) || length(by) == 0)) {
+    cli::cli_abort(
+      c(
+        "The {.arg by} argument was not supplied.",
```
dragosmg: I think it's a bit late to message about `by` not being supplied here (given there is already a processing of whatever the user supplied - i.e. `by <- .process_by(x, by)`)

R/shuffle_card.R
dragosmg: if `x` is a `bind_ard` then you have already NULL-fied `ard_args$by` so `!is.null(ard_args$by)` will always be `FALSE`.

alanahjonas95: oops yeah very good point!! , I've moved this up let me know what you think

R/shuffle_card.R
```diff
@@ -197,16 +198,17 @@ shuffle_card <- function(x,
...
     if (!is.null(ard_args$by) && !identical(ard_args$by, by)){
       cli::cli_inform(
-        "Mismatch between attributes of {.arg x} and supplied value to \\
-        {.arg by}. Attributes will be used in lieu of {.arg by}",
-        env = rlang::caller_env())
-    } else {
-      ard_args$by <- by
+        c("Mismatch between attributes of {.arg x} and supplied value to {.arg by}.",
+          "i" = "Supplied value will be used in lieu of attributes.",
+          "*" = "Note: As of {.pkg tfrmt} v0.4.0, explicitly supplied {.arg by} values take priority over data frame attributes."),
```
dragosmg:
```suggestion
          "*" = "Note: As of {.pkg tfrmt} v0.4.0, explicitly supplied {.arg by} values take priority over the {.cls card} attributes."),
```

R/shuffle_card.R
dragosmg:
```suggestion
#' @param by Grouping variable(s) used in calculations. 
```

R/shuffle_card.R
```diff
@@ -190,23 +193,30 @@ shuffle_card <- function(x,
...
+.process_by <- function(x, by) {
+  by_arg <- get_card_attr_arg(x, "by")
 
-  ard_attributes <- attributes(x)
-  ard_args <- ard_attributes$args
-  if (!is.null(by)){
-    if (!is.null(ard_args$by) && !identical(ard_args$by, by)){
+  # 1. If 'by' is explicitly supplied, it takes absolute priority
+  if (!is.null(by)) {
+    # Only warn if an attribute actually exists AND it doesn't match
```
dragosmg:
```suggestion
    # Only inform if an attribute actually exists AND it doesn't match
```

dragosmg (commented):
Looks great, thanks. Just a couple of comments + a suggestion to add a test file for `utils_cards.R`. It would help ensure the cards helpers work as expected.

R/shuffle_card.R
```diff
@@ -44,31 +44,34 @@ shuffle_card <- function(x,
...
                          fill_hierarchical_overall = "Any {colname}") {
 
-  if (!requireNamespace("cards", quietly = TRUE)) {
-    cli::cli_abort(
-      "The {.pkg cards} package must be installed to use {.fn shuffle_card}."
-    )
-  }
+  rlang::check_installed(
+    pkg = "cards",
+    version = "0.8.1",
```
bzkrouse: why do we need this since cards is a Imports in DESCRIPTION?

DESCRIPTION
```diff
@@ -42,7 +42,7 @@ Imports:
     tidyr,
     tidyselect
 Suggests:
-    cards (>= 0.8.0),
+    cards (>= 0.8.1),
```
bzkrouse: 0.8.1 provides the `bind_ard` class but i wouldnt say its a hard requirement for our code. right?

R/shuffle_card.R
```diff
@@ -44,31 +44,34 @@ shuffle_card <- function(x,
...
-  # If a combined ARD is passed, drop stale attributes and evaluate structurally
-  if (inherits(x, "bind_ard") && !is.null(attr(x, "args"))) {
-      attr(x, "args")$by <- NULL
-      attr(x, "args")$variable <- NULL
-      attr(x, "args")$strata <- NULL
+  # Check if a 'by' variable is available
+  if (is_bind_ard_card(x) && rlang::is_empty(by)) {
+    cli::cli_inform(
+      c(
+        "The {.arg by} argument was not supplied.",
```
bzkrouse:
```suggestion
        "The {.arg by} argument was not supplied and cannot be inferred from objects of class {.cls bind_ard}.",
```

R/shuffle_card.R
```diff
@@ -44,31 +44,34 @@ shuffle_card <- function(x,
...
+        "i" = "When using a combined ARD ({.cls bind_ard}) individual grouping attributes are dropped.",
```
bzkrouse: this line not needed if you apply the above change

R/shuffle_card.R
```diff
@@ -186,27 +189,34 @@ shuffle_card <- function(x,
...
 #' @noRd
-.process_by <- function(x, by){
+.process_by <- function(x, by) {
+  by_arg <- get_card_attr_arg(x, "by")
 
-  ard_attributes <- attributes(x)
-  ard_args <- ard_attributes$args
-  if (!is.null(by)){
-    if (!is.null(ard_args$by) && !identical(ard_args$by, by)){
+  # 1. If 'by' is explicitly supplied, it takes absolute priority
```
bzkrouse, line 198: add an entry to the NEWS on this

tests/testthat/test-shuffle_card.R
```diff
@@ -74,15 +74,29 @@ test_that("shuffle_card notifies user about warnings/errors before dropping", {
   )
 })
 
-test_that("shuffle_card fills missing group levels if the group is meaningful", {
+test_that("shuffle_card throws an error if bind_ard is used without a supplied by argument", {
```
bzkrouse:
```suggestion
test_that("shuffle_card message if bind_ard is used without a supplied by argument", {
```

tests/testthat/test-shuffle_card.R
```diff
@@ -74,15 +74,29 @@ test_that("shuffle_card notifies user about warnings/errors before dropping", {
...
+  # Create a combined ARD
+  combined_ard <- cards::bind_ard(
+    cards::ard_continuous(cards::ADSL, by = "ARM", variables = "AGE", statistic = ~ cards::continuous_summary_fns("mean")),
+    dplyr::tibble(group1 = "ARM", variable = "AGE", stat_name = "p", stat_label = "p", stat = list(0.05))
+  )
+
+  # Verify that the function throws the exact expected error message
```
bzkrouse:
```suggestion
  # Verify that the function throws the exact expected message
```

tests/testthat/test-shuffle_card.R
```diff
@@ -74,15 +74,29 @@ test_that("shuffle_card notifies user about warnings/errors before dropping", {
...
+test_that("shuffle_card correctly handles a combined ARD when by is explicitly supplied", {
   # mix of missing/nonmissing group levels present before shuffle
   expect_snapshot(
     cards::bind_ard(
       cards::ard_continuous(cards::ADSL, by = "ARM", variables = "AGE", statistic = ~ cards::continuous_summary_fns("mean")),
       dplyr::tibble(group1 = "ARM", variable = "AGE", stat_name = "p", stat_label = "p", stat = list(0.05))
     ) |>
       dplyr::filter(dplyr::row_number() <= 5L) |>
-      shuffle_card()
+      shuffle_card(by="ARM")
```
bzkrouse: is this change needed? i.e. did behavior of shuffle_card actually change in this case? i think it's just the message. in which case I wouldn't alter the test

bzkrouse (changes requested):
great job!! couple comments for you! 
Definitely needs a NEWS entry as well :)

DESCRIPTION
alanahjonas95: yeahh no your're right its not i will convert to 0.8.0

R/shuffle_card.R
alanahjonas95: cards is in the suggests section of the description? not imports? should i move to imports instead what do you think?

tests/testthat/test-shuffle_card.R
alanahjonas95: true just the message ive reverted the change and accepted the changed snapshot

R/shuffle_card.R
alanahjonas95: removed it

R/shuffle_card.R
alanahjonas95: added to news entry

R/shuffle_card.R
bzkrouse: oops!!! This makes total sense then ;) Just need to match the version here to the description.

NEWS.md
```diff
@@ -11,7 +11,7 @@
 * Add markdown processing of stub column labels (#617)
 * Update default for `row_grp_plan()`/`row_grp_structure()` to remove trailing post-space rows (#630, @alanahjonas95).
 * Added `extract_data()` function to enable data to be easily extracted from a tfrmt into a data frame.(#628, @alanahjonas95)
-
+* `shuffle_card()` now explicitly handles `bind_ard` attributes, prioritising the user-supplied `by` argument over inherited ARD attributes and providing clearer warning messages during mismatches (#650, @alanahjonas95).
```
bzkrouse: this feels like maybe two different NEWS items, one about ignoring ARD attributes from bind_ard objects and one about prioritizing the `by` over attributes. can you split it so they are more clear? Also, no need to add your name to the news items since you're an author :)

R/shuffle_card.R
alanahjonas95: cool ive updated the version :)

NEWS.md
alanahjonas95: let me know what you think of the updated text now, removed name from this

dragosmg (approved):
Looks great! Thanks!

bzkrouse (approved):
🥳

PR #652, opened 2026-07-08
add a covr2gh github action workflow

### LINKED ISSUE #623: enable code coverage PR comments
It would be useful to see code coverage information in the PR (a sticky comment). This can be accomplished in a couple of different ways:
* edit the codecov.io config file and set `comment: true`, or
* (shameless plug) use {covr2gh} (when a bit more stable) to have a GHA workflow that does not depend on a 3rd party service or action. A benefit of this would be that we might have a bit more control about what and how is included in the comment.

dragosmg:
This PR adds a {covr2gh} GitHub Action workflow that creates a small coverage report and posts it as a PR comment.

fixes #623


alanahjonas95 (approved):
looks great thanks for adding! do we still need the remotes for cards?

PR #655, opened 2026-07-08
make the message more helpful

bzkrouse:
Because `ard_stack` uses `bind_ard` under the hood, we are now dropping its attributes and not using them for shuffle_card (an oversight in the cards update). This is a regression, since we used to preserve the attributes of `ard_stack` and `by` was not required in `shuffle_card`. Adding a more explicit message and suggestion to the user to assist in updating their code to match previous behavior.

Note that `ard_stack_hierarchical` is not impacted.

### DIFF
--- a/R/shuffle_card.R
+++ b/R/shuffle_card.R
@@ -60,12 +60,22 @@ shuffle_card <- function(x,
     )
   }
 
-  # Check if a 'by' variable is available
+
+  # Check if a 'by' variable is available for bind_ard objs
+  by_arg <- get_card_attr_arg(x, "by")
   if (is_bind_ard_card(x) && rlang::is_empty(by)) {
+
+    by_msg <- if (!rlang::is_empty(by_arg)) {
+      c("*" = "A {.arg by} value of {.val {by_arg}} was found in the input object's attributes.",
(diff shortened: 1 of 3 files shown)


alanahjonas95 (approved):
thanks for the update to the message

R/shuffle_card.R
```diff
@@ -60,12 +60,22 @@ shuffle_card <- function(x,
...
+
+  # Check if a 'by' variable is available for bind_ard objs
+  by_arg <- get_card_attr_arg(x, "by")
   if (is_bind_ard_card(x) && rlang::is_empty(by)) {
+
+    by_msg <- if (!rlang::is_empty(by_arg)) {
+      c("*" = "A {.arg by} value of {.val {by_arg}} was found in the input object's attributes.",
+        "*" = "To use it as a grouping variable, pass it explicitly: {.code shuffle_card(by = \"{by_arg}\")}.")
+    } else {
+      c("*" = "If you want to use a grouping variable, pass it explicitly via the {.arg by} argument.")
```
dragosmg, line 72: This line is not covered by a unit test. Should we add one?

dragosmg (commented):
Just a small comment.

R/shuffle_card.R
alanahjonas95: I've added a test for this line

dragosmg (approved):
Looks great! Thanks.

PR #656, opened 2026-07-13
use air

### LINKED ISSUE #622: use Air
We might not be ready to re-format the entire codebase, but having an `air.toml` file and project-level Air settings would mean that folks using Positron or RStudio with Air will not accidentally reformat code.  

All that it's needed is a call to `usethis::use_air()` which automatically creates the required scaffolding. Additionally, we'd want `editor.formatOnSave = false` (until we are ready to switch it on).

For more details https://usethis.r-lib.org/reference/use_air.html.

dragosmg:
This PR adds the scaffolding needed to start using Air and reformats all `.R` files. 

* an empty `air.toml` file, and
* project-specific Positron / VSCode settings (sets `"editor.formatOnSave": true` to enable Air auto-formatting)
* adds a GHA workflow to check all incoming contributions are formatted correctly. the workflow fails if any lines need formatting (highlighting them).

This PR **does not** modify any existing function definitions or tests.

fixes #622


alanahjonas95 (approved):
looks good thanks

PR #657, opened 2026-07-13
increase test coverage for R/tfrmt_checks.R

### LINKED ISSUE #601: increase unit test coverage for `R/tfrmt_checks.R`
3 possible execution paths are currently untested. 

start with:
```r
devtools::test_coverage() 
```

and investigate the output for `R/tfrmt_checks.R`.

alanahjonas95:
resolves #601
Add three tests covering previously untested paths in tfrmt_checks.R:
- Error when non-first group var in col_style_structure with non-column location
- No error when row_grp_plan location is 'column' (valid case)
- Error message includes group details with spanning location

### DIFF
--- a/R/tfrmt_checks.R
+++ b/R/tfrmt_checks.R
@@ -231,11 +231,6 @@ check_col_style_row_grp_consistency <- function(x) {
         for (cap_vars_idx in seq_along(col_align_plan_as_char)) {
             grp_in_cap <- group_as_char %in%
                 col_align_plan_as_char[[cap_vars_idx]]
-            if (length(x$col_style_plan[[cap_vars_idx]]$cols) == 0) {
-                stop(
-                    "Column element is missing from col_style_structure. Note: col here refers to the values within the column variable in your data, rather than the variable name itself"
-                )
-            }
 
             if (any(grp_in_cap[-1]) && !r_grp_plan_col_loc == "column") {
                 is_invalid_plan <- TRUE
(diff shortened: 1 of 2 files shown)


dragosmg (commented):
This statement is still not covered by unit tests: 
https://github.com/GSK-Biostatistics/tfrmt/blob/main/R/tfrmt_checks.R#L235-L237

It might be worth testing if you are in this area of the codebase.

alanahjonas95:
> This statement is still not covered by unit tests: https://github.com/GSK-Biostatistics/tfrmt/blob/main/R/tfrmt_checks.R#L235-L237
> 
> It might be worth testing if you are in this area of the codebase.

So i looked into this and the the `stop()` on lines 235-237 is actually unreachable through normal usage. The `col_style_structure()` constructor (in `col_style_plan.R` line 107) already validates that col is provided and aborts before the object is ever created. https://github.com/GSK-Biostatistics/tfrmt/blob/6adafe44a81525afecd8ec0d36fd0b16a5bbee19/R/col_style_plan.R#L107 So when a `col_style_structure` ends up inside a tfrmt, its `$cols` is guaranteed to be non-empty.

This means the `if(length(...$cols) == 0)` check in `tfrmt_checks.R` is defensive dead code, do you therefore think its ok to remove this?

dragosmg:
I'd say yes, if that code can never be reached we should remove it.

alanahjonas95:
> I'd say yes, if that code can never be reached we should remove it.

ive removed would you be able to review again @dragosmg

dragosmg (approved):
looks great, thanks

PR #700, opened 2026-07-20
enable `seq_linter` 

alanahjonas95:
Replace seq_len(length(x)), 1:length(x), n():1, and 1:nrow(x) with seq_along() / rev(seq_len(n())) / seq_len(nrow()) across R sources and tests.

reslolves #701


dragosmg (approved):
looks great. keep up the good work 💪🏻

PR #735, opened 2026-07-21
734 namespaced calls

### LINKED ISSUE #734: replace importFrom directives with namespaced calls

dragosmg:
This PR:

* namespaces most of the calls from other packages (with a few exceptions)
* enables parallel testing with testthat (`Config/testthat/parallel: true`)
    * the number of cores can be controlled with an environment variable (e.g. `TESTTHAT_CPUS=7`)  
* solves 2 documentation issues:
    * exports the following internal S3 methods (these should be exported - see https://roxygen2.r-lib.org/articles/rd-other.html#s3 ("You should, however, always `@export` S3 methods, even internal ones"):
        * `eval_tidyselect_on_colvec.quosure()`
        * `eval_tidyselect_on_colvec.quosures()`
        * `expr_to_filter.quosure()`
        * `expr_to_filter.quosures()`  
    * fixes a partially documented args for an unexported function
* exceptions from namespaced calls:
    * _non-problematic_ uses:
        * {rlang} operators (`:=`,  `!!`,  `!!!`, and `%||%`)
        * the {rlang} `.data` pronoun
    * _problematic_ uses (I think these should be raised as issues):
        * {rlang} `f_rhs()` and `f_rhs<-`: we overload `f_rhs` inside `frmt_when()` (https://github.com/GSK-Biostatistics/tfrmt/issues/753)
        * {dplyr} `vars()`: a re-export of the {tidyselect} `vars()` which has since been deprecated (https://github.com/GSK-Biostatistics/tfrmt/issues/752)
  
An unrelated problematic aspect: the use of `.data` as object name. I think we should restrict this to _only_ mean {rlang}'s `.data` pronoun. (https://github.com/GSK-Biostatistics/tfrmt/issues/755)

There are a couple of implicit agreements this PR relies on. Going forward we:
* favour of namespaced calls (`package::function()`) and we only use non-namespaced calls for operators (or other relevant exceptions).
* no more `@importFrom` directives in function files. All should be centralised in the package doc. 

fixes #734


alanahjonas95 (dismissed):
thanks makes sense/ looks good :)

dragosmg:
@bzkrouse Do you think you might have time for a review? It's getting trickier to keep this PR conflict free.

dragosmg:
closing this PR as it's requiring too much effort to keep it conflict free.

PR #800, opened 2026-08-10
Convert error, warning, and message calls to cli messaging

### LINKED ISSUE #748: Harmonise condition handling and messaging
At present the package uses a mixture of base R (e.g. `stop()`), {rlang} (`abort()`) and {cli} (`cli_abort()`). I suggest we choose one and stick with it. I propose {cli}.

alanahjonas95:
resolves https://github.com/GSK-Biostatistics/tfrmt/issues/748


R/apply_footnote_meta.R
```diff
@@ -137,7 +137,7 @@ get_col_loc <- function(footnote_structure, .data, col_plan_vars, columns) {
                 )
             }
 
-            message(paste(message_text, collapse = "\n"))
+            cli::cli_inform(paste(message_text, collapse = "\n"))
```
dragosmg: Not sure the `paste()` call is needed here. The formatting seems to be off a bit (no space between `footnote` and `"Test footnote 1"` and the line break after `for the`).

We could get a bit more sophisticated and build the message (or the message details) with bullets:
``` r
cli::cli_inform(
    c(
        "The provided column location does not exist in the provided data for the footnote\"Test footnote 1\"",
        "i" = "Provided column location:",
        "*" = "trt: `Not A Column`"

    )
)
#> The provided column location does not exist in the provided data for the
#> footnote"Test footnote 1"
#> ℹ Provided column location:
#> • trt: `Not A Column`
```

R/apply_frmt_methods.R
```diff
@@ -199,7 +199,7 @@ apply_frmt.frmt_combine <- function(
         setdiff(fmt_param_vals_uq, .)
 
     if (length(miss_param_from_data) > 0) {
-        stop(paste0(
+        cli::cli_abort(paste0(
```
dragosmg: {cli} supports string interpolation with {glue}. Looking into this I realised this `stop()` / `cli::cli_abort()` call is not covered by unit tests.

R/apply_frmt_methods.R
```diff
@@ -231,7 +231,7 @@ apply_frmt.frmt_combine <- function(
     if (nrow(.tmp_data_wide) == nrow(.tmp_data)) {
         id_cols <- .tmp_data %>%
             dplyr::select(!!!column, !!label, !!!group, !!param)
-        warning(paste0(
+        cli::cli_warn(paste0(
```
dragosmg: cli + paste or paste0 can usually be substituted with cli + implicit glue interpolation

R/apply_footnote_meta.R
alanahjonas95: agreed, this is nicer. I've updated the message using cli_inform() with an "i" info line and dynamically generated "*" bullets for each column/value pair

R/apply_frmt_methods.R
alanahjonas95: Converted this to cli/glue implicit interpolation and added a new unit test

R/apply_frmt_methods.R
alanahjonas95: replaced the paste/paste0 concatenation with implicit glue interpolation. Also swept the rest of the package for any other cli + paste/paste0 call sites and converted those to implicit glue interpolation too.

PR #828, opened 2026-08-20
enable `namespace_linter`

### LINKED ISSUE #827: `namespace_linter`

alanahjonas95:
i thought we could turn this on after all your namespacing work! but the linter is flagging these starts_with function calls:

image

but i notice there are a lot of uses of starts_with within col_plan in tfrmt where it doesn't work when i add tidyselect::

so not sure what we should do about this?
shall i remove tidyselect:: as the linter suggests?

resolves #827


dragosmg:
@alanahjonas95 I'd say let's remove `tidyselect::` for now. The issues with `starts_with()` are captured in `tfrmt-package.R`:

```r
# `tidyselect::starts_with()`` is also a bit problematic since tfrmt supports
# tidyselect specification of selected columns, but the implementation has not
# kept up with subsequent tidyselect changes
```

Maybe add to this comment something about the lint.

alanahjonas95:
> @alanahjonas95 I'd say let's remove `tidyselect::` for now. The issues with `starts_with()` are captured in `tfrmt-package.R`:
> 
> ```r
> # `tidyselect::starts_with()`` is also a bit problematic since tfrmt supports
> # tidyselect specification of selected columns, but the implementation has not
> # kept up with subsequent tidyselect changes
> ```
> 
> Maybe add to this comment something about the lint.

I've updated would you be able to take another look @dragosmg

PR #831, opened 2026-08-21
enable `unnecessary_concatenation_linter`

### LINKED ISSUE #832: `unnecessary_concatenation_linter`

alanahjonas95:
resolves #832


R/mock_tbl.R
dragosmg: I think having `NULL` as default for `dont_inc` looks a bit more elegant to me.

R/mock_tbl.R
```diff
@@ -233,7 +233,7 @@ make_col_df <- function(
                 purrr::keep(is.list) %>%
                 purrr::map_dfr(function(x) {
                     span_df <- x %>%
-                        purrr::map(clean_col_names, c()) %>%
+                        purrr::map(clean_col_names, NULL) %>%
```
dragosmg: Seehttps://github.com/GSK-Biostatistics/tfrmt/pull/831/changes#r3842215749. If we do that, then there is no need to explicitly pass `NULL` here. 
```suggestion
                        purrr::map(clean_col_names) %>%
```

dragosmg (commented):
Looks great. A couple of comments.

R/mock_tbl.R
alanahjonas95: Sure ive updated this

dragosmg (approved):

